PR #25, opened 2014-07-27
Just added the "Headers" option

KyuzoR:
Just added the "Headers" option so that cookies and other various headers can be added to the request
Ex.:
Headers=Cookie: session=1234567890


idefy:
Would be nice to have this PR merged.

brianferguson:
@KyuzoR @Wallboy This has been added.

Thanks guys! (Sorry it took so long!)

-Brian

PR #29, opened 2014-09-01
GetOption() can now get values inherited from MeterStyles.

Wallboy:
The GetOption() function in Lua was unable to get any options that were inherited from meterstyles. This commit fixes that.


poiru:
This has already been implemented separately.

PR #36, opened 2015-01-25
MathParser: added Atan2 function

dcgrace:
Added standard atan2 (inputs x and y coords, returns angle rotation
around the Z axis)

Note: in testing this I discovered what looks like a fairly major bug in how the parser handles negative numbers, but I'll save that for a separate checkin.


dcgrace:
Whoops - found and fixed the bug - let me re-submit.

PR #45, opened 2015-08-24
Fixed bug translation

gmikhail:
STR_CLICKTOMANAGE, old value - "скиниамы", new value - "скинами"


poiru:
Committed as a258412. Thanks you!

PR #46, opened 2015-08-24
Fixed button ID_STR_COPYTOCLIPBOARD width

gmikhail:
In the Russian localization string does not intermeddle in the button width. The width was increased from 25 to 35.


poiru:
Thanks for the patch and sorry for the delay. Committed this as acc6f60.

PR #47, opened 2015-08-27
New meters

marcinaj:
There's an image and sample config in the commit comments.


brianferguson:
I am closing this since we switched over to D2D instead of using GDI+.

-Brian

PR #48, opened 2015-10-12
AudioLevel: Waveform Display

dcgrace:
Adds support for plugin-supplied GdiPlus::Bitmap data for use by Image meters.

Adds bitmap export to the AudioLevel plugin with variable zoom.

Note: Image meters using plugin-supplied Bitmap data bypass the Tint and ColorMatrix processing.  (I figured that stuff would be expensive to process every frame)


dcgrace:
I'm not sure why there are so many commits included - only the 7 files changed.  I think I must have done something wrong when syncing up my local version to the head Rainmeter version.

PR #51, opened 2015-10-19
AudioLevel Waveform Support

dcgrace:
- Adds bitmap-enabled plugin interface
- Adds external bitmap support to MeterImage
- Implemented AudioLevel bitmap export
- Fixed issue requiring plugin to supply PARGB data
- Fixed issue preventing scroll from working (finally!)
- New gradient colors
- Cursor color is now scriptable


dcgrace:
Whoops - found a crash bug, will update with fix.

PR #52, opened 2015-10-19
AudioLevel Waveform Support

dcgrace:
- Adds bitmap-enabled plugin interface
- Adds external bitmap support to MeterImage
- Implemented AudioLevel bitmap export
- Fixed issue requiring plugin to supply PARGB data
- Fixed issue preventing scroll from working (finally!)
- New gradient colors
- Cursor color is now scriptable
- Fixed crash on shutdown of audio device

### DIFF
--- a/Library/Measure.h
+++ b/Library/Measure.h
@@ -14,6 +14,8 @@
 #include "IfActions.h"
 #include "Util.h"
 #include "Section.h"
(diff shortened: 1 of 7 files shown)


brianferguson:
I am closing this for now since the team cannot agree if this is the right fit for measure plugins.

-Brian

Issue #53, opened 2015-11-08
WebParser can't open file if it's used by another process.


tomkort:
It appears that Wininet can't open a file if it is used by another process. For example when GPU-Z is running, it keeps updating log file at every second, it is not possible to open that file for reading with WebParser plugin. I tested fixing the bug by resorting to std::fstream if the url starts with "file://" and using PathCreateFromUrlA to easily change the url format and it seems to work.

poiru:
Fixed by 262bda7.

PR #54, opened 2015-11-08
Fixes a bug in WebParser when local file can't be opened if it's used by another process.

tomkort:
Fixes issue #53 that I filed. Uses fstream from standard library if URL points to local file. URL is converted to path with PathCreateFromUrlA so documentation can stay intact.


poiru:
@tomkort Kiitti pätchistä :+1: Lisäsin muutaman kommentin, ilmoittele jos on jotain epäselvää.

tomkort:
Eipä kestä! Fixed the issues you pointed out as I understood them. If I missed something, please let me know.

poiru:
@tomkort I committed this as 262bda7 with a few minor tweaks. Thank you!

(Fixes #54.)

PR #55, opened 2015-11-12
Create FileUtil header for common functions when handling files.

tomkort:
Created FileUtil header and implementation file. Added ReadFullFile as suggested [here](https://github.com/rainmeter/rainmeter/compare/master...tomkort:FileUtils_patch?expand=1#commitcomment-14331573).

### DIFF
--- a/Common/FileUtil.cpp
+++ b/Common/FileUtil.cpp
@@ -1,0 +1,45 @@
+﻿/* Copyright (C) 2015 Rainmeter Project Developers
+ *
+ * This Source Code Form is subject to the terms of the GNU General Public
+ * License; either version 2 of the License, or (at your option) any later
+ * version. If a copy of the GPL was not distributed with this file, You can
+ * obtain one at <https://www.gnu.org/licenses/gpl-2.0.html>. */
+
+#include "StdAfx.h"
+#include "FileUtil.h"
+
+namespace FileUtil {
+
+std::unique_ptr<BYTE[]> ReadFullFile(const std::wstring& path, size_t* size)
+{
+	FILE* file = _wfopen(path.c_str(), L"rb");
+	if (!file)
+	{
+		return nullptr;
+	}
+
+	fseek(file, 0, SEEK_END);
+	size_t fileSize = ftell(file);
+	rewind(file);
+
+	std::unique_ptr<BYTE[]> buffer(new (std::nothrow) BYTE[fileSize + 3]);
+	if (buffer)
+	{
+		fread(buffer.get(), 1, fileSize, file);
+
+		// Triple null terminate the buffer.
+		buffer[fileSize] = 0;
+		buffer[fileSize + 1] = 0;
+		buffer[fileSize + 2] = 0;
+	}
+	fclose(file);
+
+	if (size)
+	{
+		*size = fileSize;
+	}
+
+	return buffer;
+}
+
+} // namespace FileUtil
--- a/Common/FileUtil.h
+++ b/Common/FileUtil.h
@@ -1,0 +1,24 @@
+﻿/* Copyright (C) 2015 Rainmeter Project Developers
+ *
+ * This Source Code Form is subject to the terms of the GNU General Public
+ * License; either version 2 of the License, or (at your option) any later
+ * version. If a copy of the GPL was not distributed with this file, You can
+ * obtain one at <https://www.gnu.org/licenses/gpl-2.0.html>. */
+
+#ifndef RM_COMMON_FILEUTIL_H_
+#define RM_COMMON_FILEUTIL_H_
+
+#include <string>
+#include <memory>
+
+namespace FileUtil {
+
+/*
+** Reads and allocates memory for file in path. Returns unique_ptr containing addres to
+** allocated memory.
+*/
+std::unique_ptr<BYTE[]> ReadFullFile(const std::wstring& path, size_t* size = nullptr);
+
+} // namespace FileUtil
+
+#endif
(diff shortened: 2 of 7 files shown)


poiru:
@tomkort Could you rebase this on top of our `master` branch and squash your commits into one a single commit? I'll be happy to help if needed.

tomkort:
Sorry, I'm not familiar enough with git to do that. Can you give me directions?

poiru:
@tomkort Sure. Would you be able to join the [#rainmeter](https://webchat.freenode.net/?channels=#rainmeter) IRC channel on freenode? If not, I can explain here.

Plugins/PluginWebParser/WebParser.cpp
```diff
@@ -1511,28 +1512,12 @@ BYTE* DownloadUrl(HINTERNET handle, std::wstring& url, DWORD* dataSize, bool for
...
-		if (buffer)
-		{
-			fread(buffer, 1, *dataSize, file);
-
-			// Triple null terminate the buffer.
-			buffer[*dataSize] = 0;
-			buffer[*dataSize + 1] = 0;
-			buffer[*dataSize + 2] = 0;
-		}
+		// ReadFullFile returns unique_ptr. Raw BYTE* pointer is expected down the line
```
poiru: @tomkort This comment is true, but obvious from the context. Could you remote it please?

Common/FileUtil.cpp
```diff
@@ -0,0 +1,39 @@
...
+		if (!file)
+		{
+			return nullptr;
+		}
+
+		fseek(file, 0, SEEK_END);
+		DWORD dataSize = ftell(file);
+		rewind(file);
+
+		std::unique_ptr<BYTE[]> buffer(new BYTE[dataSize + 3]);
```
poiru: Can you use `new (std::nothrow) BYTE[dataSize + 3]` here?

Common/FileUtil.cpp
```diff
@@ -0,0 +1,39 @@
+﻿/* Copyright (C) 2015 Rainmeter Project Developers
+*
```
poiru: There should be a space before the asterisk.

Common/FileUtil.h
```diff
@@ -0,0 +1,23 @@
+﻿/* Copyright (C) 2015 Rainmeter Project Developers
+*
+* This Source Code Form is subject to the terms of the GNU General Public
```
poiru: There should be a space before the asterisks here.

Common/FileUtil.cpp
```diff
@@ -0,0 +1,39 @@
...
+* This Source Code Form is subject to the terms of the GNU General Public
+* License; either version 2 of the License, or (at your option) any later
+* version. If a copy of the GPL was not distributed with this file, You can
+* obtain one at <https://www.gnu.org/licenses/gpl-2.0.html>. */
+
+#include "StdAfx.h"
+#include "FileUtil.h"
+
+namespace FileUtil
+{
```
poiru: Can you change this to `namespace FileUtil {` and remove the indentation for the namespace block?

Common/FileUtil.h
```diff
@@ -0,0 +1,23 @@
...
+* obtain one at <https://www.gnu.org/licenses/gpl-2.0.html>. */
+
+#ifndef RM_COMMON_FILEUTIL_H_
+#define RM_COMMON_FILEUTIL_H_
+
+#include <string>
+#include <memory>
+
+namespace FileUtil
+{
```
poiru: Can you change this to `namespace FileUtil {` and remove the indentation for the namespace block?

Plugins/PluginWebParser/WebParser.cpp
```diff
@@ -1511,28 +1512,9 @@ BYTE* DownloadUrl(HINTERNET handle, std::wstring& url, DWORD* dataSize, bool for
...
-		{
-			fread(buffer, 1, *dataSize, file);
-
-			// Triple null terminate the buffer.
-			buffer[*dataSize] = 0;
-			buffer[*dataSize + 1] = 0;
-			buffer[*dataSize + 2] = 0;
-		}
+		BYTE* buffer = FileUtil::ReadFullFile(path).release();
+		*dataSize = strlen((char*)buffer);
```
poiru: This is problematic if the file contains embedded NULL characters (because `strlen` stops at the first NULL). Could you instead add an optional `size` parameter to `ReadFullFile`? That is, something like `std::unique_ptr<BYTE[]> ReadFullFile(const std::wstring& path, size_t* size = nullptr)`.

tomkort:
I added the optional size pointer parameter typed as `size_t`. I had to introduce temporary variable for the size since typecasting `DWORD*` to `size_t*` and using it corrupts the stack somehow... Maybe because they are not the same size: unsigned long vs unsigned long long? As `DWORD` is used elsewhere in the code to hold sizes, maybe the parameter should be also `DWORD*` even though `size_t*` makes more sense semantically.

Common/FileUtil.cpp
```diff
@@ -0,0 +1,40 @@
...
+std::unique_ptr<BYTE[]> ReadFullFile(const std::wstring& path, size_t* size)
+{
+	FILE* file = _wfopen(path.c_str(), L"rb");
+	if (!file)
+	{
+		return nullptr;
+	}
+
+	fseek(file, 0, SEEK_END);
+	*size = ftell(file);
```
poiru: Since `size` is optional, it may be null. It might be easiest to do something like:
```cpp
size_t fileSize = ftell(file);
...
...
if (size)
{
	*size = fileSize;
}

return buffer;
```

poiru:
Thanks for the patch! If you are interested, `Plugins/PluginQuote/Quote.cpp` could use `ReadFullFile` as well.

PR #56, opened 2015-11-16
Changed PluginQuote to use FileUtils::ReadFullFile.

tomkort:
As suggested [here](https://github.com/rainmeter/rainmeter/pull/55#issuecomment-156721889). Original implementation used to buffer the file and operate on it in chunks saving some memory on large files. This patch loads whole file in the buffer and so if used with large files causes increase in memory consumption.

### DIFF
--- a/Plugins/PluginQuote/PluginQuote.vcxproj
+++ b/Plugins/PluginQuote/PluginQuote.vcxproj
@@ -28,10 +28,12 @@
     <ResourceCompile Include="PluginQuote.rc" />
   </ItemGroup>
   <ItemGroup>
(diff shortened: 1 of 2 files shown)


poiru:
Oh, you're absolutely right. I misremebered that the Quote plugin was limited to 4KB, but it turns out it just uses a 4KB buffer.

Your patch looks good, but I'm going to have to close this because of my mistake. Sorry for wasting your time.

Are you interested in working on something else? I will be happy to find a bug or feature suitable for you.

tomkort:
No problem. If you have any bugs or features you think I can handle, I'm happy to give it a try!

Issue #19, opened 2013-12-16
Language: Make "Custom skin actions" localizable


poiru:

poiru:
@tomkort Do you want to take a shot at this?

[Here](https://github.com/rainmeter/rainmeter/blob/f4e7eb3bc004a176952e8e125cbc5f010312e5a6/Library/ContextMenu.cpp#L542) is where we add the `Custom skin actions` menu item. For an example of how this is done properly, check the `Update available` menu item [here](https://github.com/rainmeter/rainmeter/blob/f4e7eb3bc004a176952e8e125cbc5f010312e5a6/Library/ContextMenu.cpp#L152). This requires a bit of searching through the codebase so let me know if you get stuck!

PR #57, opened 2015-11-25
Changed Skin custom context menu title localizable solving issue #19

tomkort:


poiru:
Thanks a bunch for your patch! I committed this as 8383b19.

PR #58, opened 2015-11-25
Changed LuaScript.cpp to use FileUtil::ReadFullFile.

tomkort:


Library/lua/LuaScript.cpp
```diff
@@ -25,16 +26,8 @@ bool LuaScript::Initialize(const std::wstring& scriptFile)
...
-
-	fseek(file, 0, SEEK_END);
-	const long fileSize = ftell(file);
-	BYTE* fileData = new BYTE[fileSize];
-	fseek(file, 0, SEEK_SET);
-	fread(fileData, fileSize, 1, file);
-	fclose(file);
-	file = nullptr;
+	size_t fileSize = 0;
+	auto fileData = FileUtil::ReadFullFile(scriptFile, &fileSize);
```
poiru: We should check if `fileData` is valid (like `if (!file) return false;` in the removed code).

poiru:
@tomkort Awesome, thanks!

By the way, GitHub doesn't notify people when you force push to a PR branch so be sure to always say "Done" or something. Otherwise I might miss it.

tomkort:
Ah, sorry! I'll keep that in mind.

PR #72, opened 2016-06-06
Recreated MeterVector

TheAzack9:
Recreated meter with basic functions
Made certain that all General Meter Options works
Added function for measures


Common/Gfx/Shape.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
```
poiru: Can you remove this? `#pragma once` and the `#ifndef RM_GFX_SHAPE_H_` check below are equivalent. We use the latter.

poiru: Please add the copyright header here.

Library/MeterGeometry.h
```diff
@@ -0,0 +1,68 @@
...
+
+	virtual void Initialize();
+	virtual bool Update();
+	virtual bool Draw(Gfx::Canvas& canvas);
+
+protected:
+	virtual void ReadOptions(ConfigParser& parser, const WCHAR* section);
+	virtual void BindMeasures(ConfigParser& parser, const WCHAR* section);
+
+	//virtual bool IsFixedSize(bool overwrite = false) { return overwrite ? true : m_ImageName.empty(); }
```
poiru: Remove this.

Common/Gfx/Canvas.cpp
```diff
@@ -563,6 +563,34 @@ void Canvas::DrawMaskedBitmap(Gdiplus::Bitmap* bitmap, Gdiplus::Bitmap* maskBitm
...
+}
+
+void Canvas::DrawGeometry(const Shape & shape, D2D1_MATRIX_3X2_F & transform)
+{
+	if (!BeginTargetDraw()) return;
+
+	D2D1_MATRIX_3X2_F worldTransform;
+	m_Target->GetTransform(&worldTransform);
+	m_Target->SetTransform(transform * worldTransform);
+	Microsoft::WRL::ComPtr<ID2D1SolidColorBrush> solidBrush = nullptr;
```
poiru: No need to initialize to `nullptr`.

Common/Gfx/Shape.h
```diff
@@ -0,0 +1,20 @@
...
+#ifndef RM_GFX_SHAPE_H_
+#define RM_GFX_SHAPE_H_
+#include <gdiplus.h>
+namespace Gfx {
+	struct Shape
```
poiru: We don't indent within namespace blocks. Can you also add a newline before and after the `namespace` line?

Common/Gfx/Shape.h
```diff
@@ -0,0 +1,20 @@
...
+			m_OutlineWidth(1),
+			m_OutlineColor(Gdiplus::Color::Black)
+		{}
+
+		Microsoft::WRL::ComPtr<ID2D1Geometry> m_Shape;
+		Gdiplus::Color m_FillColor;
+		double m_OutlineWidth;
+		Gdiplus::Color m_OutlineColor;
+	};
+}
```
poiru: Please add an empty line after the `}`.

Common/Gfx/Shape.h
```diff
@@ -0,0 +1,20 @@
...
+	struct Shape
+	{
+		Shape() :
+			m_FillColor(Gdiplus::Color::White),
+			m_OutlineWidth(1),
+			m_OutlineColor(Gdiplus::Color::Black)
+		{}
+
+		Microsoft::WRL::ComPtr<ID2D1Geometry> m_Shape;
+		Gdiplus::Color m_FillColor;
```
poiru: Instead of storing `Gdiplus::Color` and then using `ToColorF`, could you just store a `D2D1_COLOR_F` directly?

Library/MeterGeometry.cpp
```diff
@@ -0,0 +1,347 @@
+#include "stdafx.h"
```
poiru: Please add the copyright header and change `stdafx` to `StdAfx`.

Library/MeterGeometry.cpp
```diff
@@ -0,0 +1,347 @@
...
+#include "MeterGeometry.h"
+#include "Logger.h"
+
+
```
poiru: Remove extra blank line.

Library/MeterGeometry.cpp
```diff
@@ -0,0 +1,347 @@
...
+MeterGeometry::MeterGeometry(Skin* skin, const WCHAR* name) : Meter(skin, name),
+	m_Shapes(),
+	m_MeasureOptions(),
+	m_NeedsRedraw(false)
+{
+}
+
+
```
poiru: Remove extra blank line.

Library/MeterGeometry.h
```diff
@@ -0,0 +1,68 @@
...
+
+private:
+	struct GeometryShape : Gfx::Shape {
```
poiru: Could this be merged into `Gfx::Shape`?

Common/Gfx/Canvas.h
```diff
@@ -74,8 +75,13 @@ class Canvas
 	void DrawMaskedBitmap(Gdiplus::Bitmap* bitmap, Gdiplus::Bitmap* maskBitmap, const Gdiplus::Rect& dstRect,
 		const Gdiplus::Rect& srcRect, const Gdiplus::Rect& srcRect2);
 
-	void FillRectangle(Gdiplus::Rect& rect, const Gdiplus::SolidBrush& brush);
+	//Needed to create shapes in meters, change it if you want. If you find a simpler way to expose this with templates etc
+	static Microsoft::WRL::ComPtr<ID2D1RectangleGeometry> CreateRectangle(D2D1_RECT_F rectangle);
```
poiru: `const D2D1_RECT_F&`, please. Can you also move this to the top (below `Finalize`) near the other static functions.

Library/MeterGeometry.h
```diff
@@ -0,0 +1,68 @@
...
+	public Meter
+{
+public:
+	MeterGeometry(Skin* skin, const WCHAR* name);
+	~MeterGeometry();
+
+	MeterGeometry(const MeterGeometry& other) = delete;
+	MeterGeometry& operator=(MeterGeometry other) = delete;
+
+	virtual UINT GetTypeID() { return TypeID<MeterGeometry>(); }
```
poiru: Can you remove `virtual` and use `override` instead?

Library/MeterGeometry.cpp
```diff
@@ -0,0 +1,347 @@
...
+Microsoft::WRL::ComPtr<ID2D1Geometry> MeterGeometry::ParseRect(GeometryShape& shape, RECT& rect)
+{
+	D2D1_RECT_F geo_rect = D2D1::RectF(rect.left, rect.top, rect.right + rect.left, rect.bottom + rect.top);
+	geo_rect.left += shape.m_OutlineWidth / 2;
+	geo_rect.top += shape.m_OutlineWidth / 2;
+	return Gfx::Canvas::CreateRectangle(geo_rect);
+}
+
+double MeterGeometry::ParseRotation(const LPCWSTR& string, double defaultValue, GeometryShape& shape)
```
poiru: Can you use `const WCHAR*` instead of `const LPCWSTR&` here and elsewhere?

Common/Gfx/Shape.h
TheAzack9: The reason i used Gdiplus::Color is because it is what ConfigParser::ParseColor returns. I could change it to D2D1_COLOR_F, but i'd still have to convert it unless i create another method to parse a D2D1_COLOR_F, which imho is not needed. :)

Library/MeterGeometry.h
```diff
@@ -0,0 +1,68 @@
+#pragma once
```
poiru: Comments for `Shape.h` apply here, too.

Library/MeterGeometry.h
```diff
@@ -0,0 +1,68 @@
...
+#ifndef __METERGEOMETRY_H__
```
poiru: Use `RM_LIBRARY_METERGEOMETRY_H_`.

Library/MeterGeometry.h
```diff
@@ -0,0 +1,68 @@
...
+#include "Meter.h"
+#include <vector>
+#include "..\Common\Gfx\Shape.h"
+
+class MeterGeometry :
+	public Meter
+{
+public:
+	MeterGeometry(Skin* skin, const WCHAR* name);
+	~MeterGeometry();
```
poiru: This should be `virtual`.

Library/MeterGeometry.h
TheAzack9: It could, but i want to make Gfx::Shape only contain the fields needed to actually render the shape, and let all possible calculations (rotation, offset, skew, etc) happen in the Meter class instead of in the DrawGeometry method.

Library/MeterGeometry.h
TheAzack9: I'm guessing you mean the commented method... done 
I just forgot it existed...

Common/Gfx/Canvas.cpp
TheAzack9: I've always thought that it was a good thing to initalize it to null so that it doesn't take some garbage value. It isn't needed tho so i'll remove it.

poiru:
@TheAzack9 Thank you for the patch. I added a bunch of comments. Let me know if you have questions!

Common/Gfx/Canvas.cpp
poiru: @TheAzack9 That's true with e.g. `int`. `ComPtr` has a constructor which initializes it properly.

Common/Gfx/Shape.h
poiru: @TheAzack9 Yep, that's what I meant. Store the converted value here.

Library/MeterGeometry.h
poiru: Makes sense, thanks for the clarification!

Library/MeterGeometry.h
TheAzack9: This didn't seem to work it might be because i misunderstood something tho.

poiru: `override` goes after the `()`. See e.g. [this](https://github.com/rainmeter/rainmeter/blob/22066aaf92514fc665806f7edb733267f82fe37b/Library/MeasureRecycleManager.h#L22).

TheAzack9: TIL

Common/Gfx/Shape.h
```diff
@@ -1,20 +1,30 @@
...
+
 #include <gdiplus.h>
+
 namespace Gfx {
-	struct Shape
+
+struct Shape
+{
+	Shape() :
+		m_FillColor(D2D1::ColorF(D2D1::ColorF::White)),
```
poiru: Can this be just `m_FillColor(D2D1::ColorF::White)`?

Library/MeterGeometry.h
TheAzack9: I will probably move it to a separate header later, as i know it will get cluttered

Common/Gfx/Shape.h
```diff
@@ -1,20 +1,30 @@
-#pragma once
+/* Copyright (C) 2013 Rainmeter Project Developers
+*
+* This Source Code Form is subject to the terms of the GNU General Public
+* License; either version 2 of the License, or (at your option) any later
+* version. If a copy of the GPL was not distributed with this file, You can
+* obtain one at <https://www.gnu.org/licenses/gpl-2.0.html>. */
```
poiru: There should be a space before the `*`. VS removes annoyingly removes them... Could you also change the year to 2016?

Library/MeterGeometry.h
```diff
@@ -11,12 +10,12 @@ class MeterGeometry :
...
-	~MeterGeometry();
+	virtual ~MeterGeometry();
 
 	MeterGeometry(const MeterGeometry& other) = delete;
 	MeterGeometry& operator=(MeterGeometry other) = delete;
 
-	virtual UINT GetTypeID() { return TypeID<MeterGeometry>(); }
+	UINT GetTypeID() override { return TypeID<MeterGeometry>(); }
 
 	virtual void Initialize();
```
poiru: This and all `virtual` functions below should also use `override`.

Library/MeterGeometry.h
```diff
@@ -0,0 +1,65 @@
...
+public:
+	MeterGeometry(Skin* skin, const WCHAR* name);
+	virtual ~MeterGeometry();
+
+	MeterGeometry(const MeterGeometry& other) = delete;
+	MeterGeometry& operator=(MeterGeometry other) = delete;
+
+	UINT GetTypeID() override { return TypeID<MeterGeometry>(); }
+
+	virtual void Initialize();
```
poiru: This and all `virtual` functions below should also use `override`.

Library/MeterGeometry.h
```diff
@@ -0,0 +1,65 @@
+#ifndef RM_LIBRARY_METERGEOMETRY_H_
```
poiru: Could you add the copyright header here?

Common/Gfx/Shape.h
TheAzack9: Doesn't work sadly as D2D1::ColorF::White is just a enum value.

poiru: I see, thanks for checking!

poiru:
@TheAzack9 Thanks for the quick fixes! I have further questions/comments about the actual logic in `MeterGeometry.cpp`, but sadly I don't have time to dig into it today. Expect more comments tomorrow!

poiru:
@TheAzack9 Today turned out to be busier than expected. I'll get to this tomorrow, sorry for the delay!

TheAzack9:
I can't reply to comments for some reason, but i just wanted to reply that it's no problem.. I will be quite busy tomorrow so i can't really fix anything before around five pm (GMT + 1)

Common/Gfx/Canvas.cpp
```diff
@@ -563,6 +570,29 @@ void Canvas::DrawMaskedBitmap(Gdiplus::Bitmap* bitmap, Gdiplus::Bitmap* maskBitm
 	bitmapLock->Release();
 }
 
+void Canvas::DrawGeometry(const Shape & shape, D2D1_MATRIX_3X2_F & transform)
+{
+	if (!BeginTargetDraw()) return;
+
+	m_Target->SetAntialiasMode(D2D1_ANTIALIAS_MODE_ALIASED);
```
poiru: Could you do something like `const auto originalTransform = m_Target->GetAntialiasMode();` and use that in the `SetAntialiasMode` call at the end?

Common/Gfx/Canvas.cpp
```diff
@@ -563,6 +570,29 @@ void Canvas::DrawMaskedBitmap(Gdiplus::Bitmap* bitmap, Gdiplus::Bitmap* maskBitm
...
+	HRESULT hr = m_Target->CreateSolidColorBrush(shape.m_FillColor, solidBrush.GetAddressOf());
+	if (SUCCEEDED(hr)) {
+		if (shape.m_FillColor.a > 0)
+			m_Target->FillGeometry(shape.m_Shape.Get(), solidBrush.Get());
+
+		solidBrush->SetColor(shape.m_OutlineColor);
+		if (shape.m_OutlineColor.a > 0)
+			m_Target->DrawGeometry(shape.m_Shape.Get(), solidBrush.Get(), shape.m_OutlineWidth);
+	}
+	solidBrush.Reset();
```
poiru: Please remove this line. The destructor of `ComPtr` will automatically do this.

Common/Gfx/Canvas.cpp
```diff
@@ -563,6 +570,29 @@ void Canvas::DrawMaskedBitmap(Gdiplus::Bitmap* bitmap, Gdiplus::Bitmap* maskBitm
...
+
+	m_Target->SetAntialiasMode(D2D1_ANTIALIAS_MODE_ALIASED);
+	D2D1_MATRIX_3X2_F worldTransform;
+	m_Target->GetTransform(&worldTransform);
+	m_Target->SetTransform(transform * worldTransform);
+	Microsoft::WRL::ComPtr<ID2D1SolidColorBrush> solidBrush;
+	HRESULT hr = m_Target->CreateSolidColorBrush(shape.m_FillColor, solidBrush.GetAddressOf());
+	if (SUCCEEDED(hr)) {
+		if (shape.m_FillColor.a > 0)
+			m_Target->FillGeometry(shape.m_Shape.Get(), solidBrush.Get());
```
poiru: Please use `{ ... }` after the `if` here and below.

Common/Gfx/Canvas.cpp
```diff
@@ -563,6 +570,29 @@ void Canvas::DrawMaskedBitmap(Gdiplus::Bitmap* bitmap, Gdiplus::Bitmap* maskBitm
...
 }
 
+void Canvas::DrawGeometry(const Shape & shape, D2D1_MATRIX_3X2_F & transform)
+{
+	if (!BeginTargetDraw()) return;
+
+	m_Target->SetAntialiasMode(D2D1_ANTIALIAS_MODE_ALIASED);
+	D2D1_MATRIX_3X2_F worldTransform;
+	m_Target->GetTransform(&worldTransform);
+	m_Target->SetTransform(transform * worldTransform);
```
poiru: Add a new line after this for semantic separation.

Common/Gfx/Canvas.cpp
```diff
@@ -563,6 +570,29 @@ void Canvas::DrawMaskedBitmap(Gdiplus::Bitmap* bitmap, Gdiplus::Bitmap* maskBitm
...
+	Microsoft::WRL::ComPtr<ID2D1SolidColorBrush> solidBrush;
+	HRESULT hr = m_Target->CreateSolidColorBrush(shape.m_FillColor, solidBrush.GetAddressOf());
+	if (SUCCEEDED(hr)) {
+		if (shape.m_FillColor.a > 0)
+			m_Target->FillGeometry(shape.m_Shape.Get(), solidBrush.Get());
+
+		solidBrush->SetColor(shape.m_OutlineColor);
+		if (shape.m_OutlineColor.a > 0)
+			m_Target->DrawGeometry(shape.m_Shape.Get(), solidBrush.Get(), shape.m_OutlineWidth);
+	}
```
poiru: Add a new line after this for semantic separation.

Common/Gfx/Canvas.h
```diff
@@ -20,6 +20,7 @@
 #include <dwrite_1.h>
 #include <wincodec.h>
 #include <wrl/client.h>
+#include "Shape.h";
```
poiru: Please remove the `;` and move this above the system headers.

Common/Gfx/Canvas.h
```diff
@@ -74,8 +76,10 @@ class Canvas
 	void DrawMaskedBitmap(Gdiplus::Bitmap* bitmap, Gdiplus::Bitmap* maskBitmap, const Gdiplus::Rect& dstRect,
 		const Gdiplus::Rect& srcRect, const Gdiplus::Rect& srcRect2);
 
-	void FillRectangle(Gdiplus::Rect& rect, const Gdiplus::SolidBrush& brush);
+	void DrawGeometry(const Shape& shape, D2D1_MATRIX_3X2_F& transform);
```
poiru: `const D2D1_MATRIX_3X2_F&` please.

Common/Gfx/Shape.h
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * This Source Code Form is subject to the terms of the GNU General Public
+ * License; either version 2 of the License, or (at your option) any later
+ * version. If a copy of the GPL was not distributed with this file, You can
+ * obtain one at <https://www.gnu.org/licenses/gpl-2.0.html>. */
+
+#ifndef RM_GFX_SHAPE_H_
+#define RM_GFX_SHAPE_H_
+
+#include <gdiplus.h>
```
poiru: This include is unnecessary, but might want to include `<d2d1_1.h>` and `<wrl/client.h>` for correctness.

Common/Gfx/Shape.h
```diff
@@ -0,0 +1,31 @@
...
+	{
+	}
+
+	Microsoft::WRL::ComPtr<ID2D1Geometry> m_Shape;
+	D2D1_COLOR_F m_FillColor;
+	double m_OutlineWidth;
+	D2D1_COLOR_F m_OutlineColor;
+};
+
+}
```
poiru: Add `}  // namespace Gfx` comment and a newline after this.

Library/MeterGeometry.cpp
```diff
@@ -0,0 +1,370 @@
...
+}
+
+void MeterGeometry::Initialize()
+{
+	Meter::Initialize();
+}
+
+bool MeterGeometry::ParseShape(GeometryShape & shape, const LPCWSTR & optionName, const LPCWSTR & optionValue)
+{
+	if (_wcsicmp(optionName, L"Rectangle") == 0) shape.m_Shape = ParseRect(shape, ConfigParser::ParseRECT(optionValue));
```
poiru: Please use

```cpp
if (...) {

} else {

}
```

Library/MeterGeometry.cpp
```diff
@@ -0,0 +1,370 @@
...
+	else return false;
+	//Easy to do with a rectangle, but this will calculate the bounds of all shape types
```
poiru: Please remove this comment.

Library/MeterGeometry.cpp
```diff
@@ -0,0 +1,370 @@
...
+	if (shape.m_Shape) {
```
poiru: Could we separate this bit out into another function like `UpdateSize()` or something?

Library/MeterGeometry.cpp
```diff
@@ -0,0 +1,370 @@
...
+		shape.m_Shape->GetBounds(D2D1::Matrix3x2F::Identity(), &shape.m_Bounds);
```
poiru: Could we have
```const auto shapeWidth = std::ceil(shape.m_Bounds.right + shape.m_OutlineWidth / 4 * 3 + GetWidthPadding());`
and then use `shapeWidth` in the `if` and the `m_W` assignment? That would avoid duplicating the expression.

File: Library/MeterGeometry.cpp [hunk 35]
```diff
@@ -0,0 +1,370 @@
...
+		if (!m_WDefined && m_W < shape.m_Bounds.right + std::ceilf(shape.m_OutlineWidth / 4 * 3) + GetWidthPadding()) {
+			m_W = std::ceil(shape.m_Bounds.right + shape.m_OutlineWidth / 4 * 3 + GetWidthPadding());
+			if (!IsHidden())
```
poiru: Is the `IsHidden` check needed?

Library/MeterGeometry.cpp
```diff
@@ -0,0 +1,370 @@
...
+			if(!IsHidden())
+				m_Skin->SetResizeWindowMode(RESIZEMODE_CHECK);
+		}
+	}
+	return true;
+}
+
+bool MeterGeometry::ReplaceShapeOption(GeometryShape & shape, const LPCWSTR & optionName, const LPCWSTR & optionValue)
+{
+	if (_wcsicmp(optionName, L"FillColor") == 0)			shape.m_FillColor = ToColorF(ConfigParser::ParseColor(optionValue));
```
poiru: Please use
```cpp
if (...) {
  return true;
} else if (...) {
  return true;
}
return false;
```

Library/MeterGeometry.cpp
```diff
@@ -0,0 +1,370 @@
...
+		m_W = 0;
+	if (!m_HDefined)
+		m_H = 0;
+
+	m_Shapes.clear();
+	m_MeasureOptions.clear();
+	size_t currentShapeId = 1;
+	std::wstring shapeName = L"Shape";
+	std::wstring shapeOption = parser.ReadString(section, L"Shape", L"");
+	std::map<LPCWSTR, std::pair<LPCWSTR, LPCWSTR>> postOptions;
```
poiru: Can you explain why `postOptions` is needed?

Library/MeterGeometry.cpp
```diff
@@ -0,0 +1,370 @@
...
+	if (!shapeOption.empty())
```
poiru: Can you simplify the `if` and `do` to just:
```
while (!shapeOption.empty())
{
}
```

Library/MeterGeometry.cpp
```diff
@@ -0,0 +1,370 @@
...
+	return false;
+}
+
+bool MeterGeometry::Draw(Gfx::Canvas & canvas)
+{
+	if (!Meter::Draw(canvas)) return false;
+	
+	for (const auto& shape : m_Shapes) {
+		if (shape.second.m_Shape) {
+			D2D1_POINT_2F centerPoint;
```
poiru: `const auto centerPoint = D2D1::Point2F(...)` please.

Library/MeterGeometry.cpp
```diff
@@ -0,0 +1,370 @@
...
+			centerPoint = D2D1::Point2F(shape.second.m_RotationCenter.x, shape.second.m_RotationCenter.y);
+			D2D1_MATRIX_3X2_F transform = D2D1::Matrix3x2F(
```
poiru: `const auto transform` please.

Library/MeterGeometry.cpp
```diff
@@ -0,0 +1,370 @@
...
+		}
+	}
+	return false;
+}
+
+bool MeterGeometry::Draw(Gfx::Canvas & canvas)
+{
+	if (!Meter::Draw(canvas)) return false;
+	
+	for (const auto& shape : m_Shapes) {
```
poiru: Can you change this to:
```
for (const auto& it : m_Shapes)
{
  const auto& shape = it.second;
}
```
and then use `shape` instead of `shape.second` below.

Library/MeterGeometry.cpp
```diff
@@ -0,0 +1,370 @@
...
+	std::vector<std::wstring> tokens;
+
+	size_t lastPos, pos = 0;
+	do
+	{
+		lastPos = str.find_first_not_of(delimiters, pos);
+		if (lastPos == std::wstring::npos) break;
+
+		pos = str.find_first_of(delimiters, lastPos + 1);
+		std::wstring token = str.substr(lastPos, pos - lastPos);  // len = (pos != std::wstring::npos) ? pos - lastPos : pos
```
poiru: Please remove all commented out code.

Library/MeterGeometry.cpp
```diff
@@ -0,0 +1,370 @@
...
+		if (shape.m_RotationCenter.y == FLT_MIN) shape.m_RotationCenter.y = (shape.m_Bounds.bottom - shape.m_Bounds.top) / 2 - shape.m_Bounds.top;
+	}
+	else {
+		shape.m_RotationCenter.x = (shape.m_Bounds.right - shape.m_Bounds.left) / 2 - shape.m_Bounds.left;
+		shape.m_RotationCenter.y = (shape.m_Bounds.bottom - shape.m_Bounds.top) / 2 - shape.m_Bounds.top;
+	}
+	return rotation;
+}
+
+std::vector<std::wstring> MeterGeometry::CustomTokenize(const std::wstring& str, const std::wstring& delimiters)
```
poiru: Can you explain how why we need this (i.e. what does this do differently compared to `ConfigParser::Tokenize`)?

Library/MeterGeometry.cpp
```diff
@@ -0,0 +1,370 @@
...
+
+		if (pos == std::wstring::npos) break;
+		if (!offsett)
+			++pos;
+	} while (true);
+
+	return tokens;
+}
+//Modified version of ParseRect
+D2D1_POINT_2F MeterGeometry::ParsePoint(const LPCWSTR& string, double defaultVal)
```
poiru: Could you replace this with a static ConfigParser function that does something like:
```
double x = defaultVal;
double y = defaultVal;
double dummy = 0.0;
ParseInt4(x, y, dummy, dummy);

File: Library/MeterGeometry.h [hunk 45]
```diff
@@ -0,0 +1,72 @@
...
+
+		double m_Rotation;
+		D2D1_POINT_2F m_RotationCenter;
+		D2D1_POINT_2F m_Skew;
+		D2D1_SIZE_F m_Scale;
+		D2D1_SIZE_F m_Offset;
+
+		D2D1_RECT_F m_Bounds;
+	};
+	bool ParseShape(GeometryShape & shape, const LPCWSTR& optionName, const LPCWSTR& optionValue);
```
poiru: Please replace all instances of `const LPCWSTR&` with `const WCHAR*`.

Library/MeterGeometry.h
```diff
@@ -0,0 +1,72 @@
...
+	void Initialize() override;
+	bool Update() override;
+	bool Draw(Gfx::Canvas& canvas) override;
+
+protected:
+	void ReadOptions(ConfigParser& parser, const WCHAR* section) override;
+	void BindMeasures(ConfigParser& parser, const WCHAR* section) override;
+
+private:
+	struct GeometryShape : Gfx::Shape {
```
poiru: Please put `{` on its own line.

Library/MeterGeometry.h
```diff
@@ -0,0 +1,72 @@
...
+		{}
+
+		double m_Rotation;
+		D2D1_POINT_2F m_RotationCenter;
+		D2D1_POINT_2F m_Skew;
+		D2D1_SIZE_F m_Scale;
+		D2D1_SIZE_F m_Offset;
+
+		D2D1_RECT_F m_Bounds;
+	};
```
poiru: An empty line after this, please.

Library/MeterGeometry.cpp
```diff
@@ -0,0 +1,370 @@
...
+				);
+			canvas.DrawGeometry(shape.second, transform);
+		}
+	}
+	return true;
+}
+/*
+** Not necessary for the Geometry meter to be bound to any measures
+**
+*/
```
poiru: Remove this comment block.

Library/MeterGeometry.cpp
```diff
@@ -0,0 +1,370 @@
...
+void MeterGeometry::BindMeasures(ConfigParser & parser, const WCHAR * section)
```
poiru: Remove space before `&` and `*`.

Library/MeterGeometry.cpp
```diff
@@ -0,0 +1,370 @@
...
+	m_Shapes.clear();
+	m_MeasureOptions.clear();
+	size_t currentShapeId = 1;
+	std::wstring shapeName = L"Shape";
+	std::wstring shapeOption = parser.ReadString(section, L"Shape", L"");
+	std::map<LPCWSTR, std::pair<LPCWSTR, LPCWSTR>> postOptions;
+	if (!shapeOption.empty()) 
+		do
+		{
+			std::vector<std::wstring> shapeTokens = CustomTokenize(shapeOption.c_str(), L"|:");
```
poiru: I'm not sure this is a good approach because it means you can mix and match `:` and `|`. Maybe you could first split by `|` and then split each of those tokens by `:`?

Common/Gfx/Canvas.cpp
TheAzack9: Suddenly did the thread expand quite a bit, i'll fix it now ;)

Yeah, this would probably be better, i simply mirrored the DrawMaskedBitmap...

Common/Gfx/Canvas.cpp
TheAzack9: I had some stupid issues with ComPtrs and memory consumption, seems the gods of ComPtrs are kind today tho (I think it was because i was allocating a new brush on top of an existing. This would cause the previous ComPtr to never reset, right?)

Common/Gfx/Canvas.h
TheAzack9: Do you mind explaining what makes const a better choice here?

Library/MeterGeometry.cpp
TheAzack9: Some shape modifiers are dependent on other shape modifiers, in this case is it Rotation which needs the width of the shape and the height of the meter to calculate the center point. The shape width and height isn't set before the shape is actually created, which happens after all the options are loaded (as the shape needs outlineWidth to calculate the correct bounds)

Library/MeterGeometry.cpp
TheAzack9: As i said in the commented parts, what i do is match some of the mathematical expressions "( with ), [ with ] and # with #" so that there are an equal amount of them inside one token. (if there isn't then merge with the next token) This fixes the issue the current Tokenize function have where for example Shape=Rectangle:(x > y ? a : b), 0, 100, 100 would split the : into different tokens, when you really only wanted to stay as a mathematical expression.

Library/MeterGeometry.cpp
TheAzack9: This is probably a better idea, i just have to make the static ParseInt4 function first ;)

Library/MeterGeometry.h
TheAzack9: again, do you mind explaining what's better with const WCHAR* over const LPCWSTR&?

Library/MeterGeometry.cpp
TheAzack9: I didn't notice this until now... well there's another weird thing Visual Studio does (when auto generating methods into the cpp file)

Library/MeterGeometry.cpp
TheAzack9: You are absolutely correct in this not being right, lol. I changed it this way because Tokenize uses quite a bit of computing power (Especially when it's run ~10 times per shape), but i do have some ideas on how to solve this tho.

Library/MeterGeometry.h
poiru: `LPCWSTR` is a typedef for `const WCHAR*`. So you basically have `const WCHAR*&`, which is a little confusing.

Common/Gfx/Canvas.h
poiru: Always use `const` if you can because it tells the reader that the variable will not be modified.

Common/Gfx/Canvas.cpp
poiru: You should use `ReleaseAndGetAddressOf()` instead of `GetAddressOf()` in that case.

Library/MeterGeometry.cpp
TheAzack9: This wasn't really a good idea, since ParseInt4 obviously casts the number to int and i need the points to be in double/float...

TheAzack9:
Closed for now!

brianferguson:
Committed with changes.
https://github.com/rainmeter/rainmeter/commit/d50fa1654255c0431115e8878b1e933a79d95df7

PR #73, opened 2016-08-27
Added support for require function

TheAzack9:
This is a small fix to make the require function fetch from the resources folder automatically.
Another folder might be preferred, but i decided upon the resources folder for ease of use.


TheAzack9:
Closed because of bleeding occuring

TheAzack9:
Did not work :C

PR #93, opened 2017-02-01
Illustro measureIP - IP from SystemInfoData

glebocki:
IP address is taken from SystemInfoData.


brianferguson:
The SysInfo plugin asks Windows for the 'internal' IP address from your local network (usually from your router). To get the 'external' IP address (the address your ISP assigns to you), you have to use a website that exposes it so that Webparser can see it.

-Brian

PR #97, opened 2017-02-16
Fix Lyrics

Alia5:
- Removed LYRDB as it doesn't exist anymore
- Updated Letras URL and fixed Stringmanip.
- Fixed inclusion of LyricWiki footer


Alia5:
It's been almost 2 years...
Haven't had time yet?!

Alia5:
That took a while...

brianferguson:
Considering it's an undocumented feature that is hardly used...not to mention the sad state most of the players are in due to the ever changing (and restrictive) access to their API's...you're lucky we even accepted this PR.

-Brian

PR #105, opened 2017-05-15
change now playing position to return double

NighthawkSLO:
this only effects the time on the number value, changing it from eg. 10 to 10.233
this would allow skin makers to have a bit more control in progress bars or similar things

### DIFF
--- a/Library/NowPlaying/Player.h
+++ b/Library/NowPlaying/Player.h
@@ -87,7 +87,7 @@ public:
 	LPCTSTR GetCoverPath() const { return m_CoverPath.c_str(); }
 	LPCTSTR GetFilePath() const { return m_FilePath.c_str(); }
 	UINT GetDuration() const { return m_Duration; }
-	UINT GetPosition() const { return m_Position; }
+	double GetPosition() const { return m_Position; }
 	UINT GetRating() const { return m_Rating; }
 	UINT GetVolume() const { return m_Volume; }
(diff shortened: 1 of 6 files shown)


brianferguson:
Unfortunately this cannot be accepted because most players only return position values in whole seconds.

Examples:
The CAD interface returns integers: https://github.com/poiru/vlc-libcad/blob/master/cad.c#L353-L358
The WMP interface returns values in whole seconds: https://www.leadtools.com/help/v18/multimedia/mediafoundation/iltmfplay__get_currentposition.html
Itunes returns the position in seconds (note the `long` integral type): https://github.com/rainmeter/rainmeter/blob/master/Library/NowPlaying/PlayerITunes.cpp#L311-L313

Winamp does return the position in milliseconds (which can be used to get a fractional position), however, given the above players do not, we will not be changing this interface.

Just changing the datatype from an integral type to a floating-point will not work in most of the these cases since the underlying players will only return the position with an integral type.

-Brian

tjhrulz:
Yeah my apologies I should have built and tested this before I told him to commit it but my visual studio build was borked. What I was hoping to achieve was to have the Progress measure and possibly SetPosition take percentages that are not just integers. That would not require the time to be in ms to have an impact as that should improve any animations that use progress and make setPosition more accurate. I still need to look at getting setPosition working and make sure Progress is correct on everything, I'll make a separate PR request once I get both tested and working with all player types.

PR #106, opened 2017-05-18
Better accuracy in NowPlaying for SetPosition and Progress

tjhrulz:
Measure Progress no longer divides Position and Duration into an integer causing an unneeded
loss of data.
SetPosition no longer reads the string as an int but instead reads it as
a float.

These changes are important to maintain accuracy in longer songs.

### DIFF
--- a/Library/MeasureNowPlaying.cpp
+++ b/Library/MeasureNowPlaying.cpp
@@ -330,7 +330,7 @@ void MeasureNowPlaying::UpdateValue()
 	case MEASURE_PROGRESS:
 		if (player->GetDuration())
 		{
-			m_Value = (player->GetPosition() * 100) / player->GetDuration();
+			m_Value = ((double)player->GetPosition() * 100) / player->GetDuration();
 		}
 		break;
 	case MEASURE_RATING:
@@ -468,7 +468,7 @@ void MeasureNowPlaying::Command(const std::wstring& command)
 
 			if (_wcsnicmp(args, L"SetPosition", 11) == 0)
 			{
-				int position = (_wtoi(arg) * (int)player->GetDuration()) / 100;
+				int position = (int)(_wtof(arg) * player->GetDuration()) / 100;
 				if (arg[0] == L'+' || arg[0] == L'-')
 				{
 					position += player->GetPosition();


brianferguson:
`GetPosition` returns values as integers, so type-casting it to a floating-point type will do nothing.

For example:
`m_Value = ((double)player->GetPosition() * 100) / player->GetDuration();`
`GetPosition` is an integral type (UINT), which is then type-cast to a floating-point type (double). The value will not change just by type-casting. If `GetPosition` returned a value like `10`, type-casting it to a double will just convert it to `10.000`. Since the underlying data type of the position is an integral type, there will never be a fractional part returned. I hope this makes sense.

Also, I hope you can see the problem here as well:
`int position = (int)(_wtof(arg) * player->GetDuration()) / 100;`
This correctly parses a floating-point number, mulitplies it by the duration, then type-casts it to a integer (losing its fractional part), then divided by a 100 and stores it in a integer. So basically it goes from a floating-point number to an integer, then stored as a integer.


What you are trying to accomplish is going to take a bit more work than just changing a data type. Changing the values from an integral type to a floating-point type is just the beginning. The underlying libraries that retrieve the actual data from the player will need to be changed - which might not be possible if the player does not expose that data as floating-point values.

-Brian

tjhrulz:
It think you are forgetting that SetPosition and Progress use a percents not a timestamps. I have tested both of these, they are both more accurate by large amount in long songs. 

For `m_Value = ((double)player->GetPosition() * 100) / player->GetDuration();` is going to give more accuracy if position is not a perfect multiple of duration / 100 because m_Value is not an int in instead of getting 98 I can get 98.705.

In the case of `int position = (int)(_wtof(arg) * player->GetDuration()) / 100;` Position and duration are times in seconds multiplying by a float is going to give you more accuracy, i.e. Lets say I have a 452 second song and I want to set it to 45.5% of the way through if I were to use the old method I would set it to 203 seconds into it. Using the new method I would correctly set it to 205 seconds.

brianferguson:
You are totally right, I was thinking that integer division was playing a roll in this when it isn't (I must be thinking of a different language).

This looks acceptable.

PR #115, opened 2017-08-02
Fix Spotify not always updating state to playing

tjhrulz:
Setting the state to playing before only happened if there was a change to the title or artist.


tjhrulz:
This fixes #115

Issue #17, opened 2013-12-16
NowPlaying: Shuffle/Repeat status not updated with iTunes


poiru:
http://rainmeter.net/forum/viewtopic.php?p=93733#p93733

Joehuu:
This has since been fixed in https://github.com/rainmeter/rainmeter/pull/123.

PR #123, opened 2017-09-02
Fixes shuffle and repeat not updating

tjhrulz:
Should fix a 4 year old bug that the repeat and shuffle state are not
update except on song change. Also removes some no longer needed code. I
think we could be more agressive and never check repeat or shuffle state
outside updateData but I left those in.


tjhrulz:
Sorry for the original bad PR. Fixes #17

PR #124, opened 2017-09-10
Plugin section variables

tjhrulz:
Should fit the latests request minus the last couple minor tweaks you wanted. 
This will allow plugins to be called inline similar to the new lua feature with the same format. Any function (except the basic rainmeter ones) that is marked to be exported by the plugin will be able to be called so once plugins implement support you will need to check out their documentation.


tjhrulz:
Added as of https://github.com/rainmeter/rainmeter/commit/220df02c8e4a84d25feaad8213c8b64fc5ddef47

PR #125, opened 2017-09-15
Hotfix for iTunes 12.7 breaking COM events

tjhrulz:
I now have to do checks if title and the like have changed so that we
are using up 0% CPU instead of 10%. A few other changes probably could
be made but with any luck Apple will realize how stupid they were to
remove this and will add the event handlers back in.


tjhrulz:
Hotfix seems to still work with iTunes 12.7.1.

PR #127, opened 2017-09-20
Changed site to get WAN IP address: icanhazip.com

### LINKED ISSUE #126: Illustro's Network skin doesn't load IP Address
It seems that get-ip.me is down.

adriancampos:
Fixes #126. Use http://icanhazip.com instead of http://get-ip.me

### DIFF
--- a/Build/Skins/illustro/Network/Network.ini
+++ b/Build/Skins/illustro/Network/Network.ini
@@ -39,9 +39,9 @@ MaxUpload=10485760
 ; Rainmeter, take a look at some of the other illustro skins before modifying this one.
 Measure=Plugin
 Plugin=WebParser
-URL=http://get-ip.me/
+URL=http://icanhazip.com/
 UpdateRate=14400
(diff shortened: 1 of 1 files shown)


MarcoPixel:
I would recommend using http://checkip.dyndns.org instead of http://icanhazip.com as it is a service provided by Oracle and thus should be more reliable in theory.

```ini
Url=http://checkip.dyndns.org
RegExp="(?siU)Address: (.*)</body>"
```

brianferguson:
Thanks. We decided to go with [http://checkip.amazonaws.com/](http://checkip.amazonaws.com/) instead. We will also be adding a link to our docs site with other options if future problems arise.

[Relevant code](https://github.com/rainmeter/rainmeter/blob/master/Build/Skins/illustro/Network/Network.ini#L37-L49).
[Commit](https://github.com/rainmeter/rainmeter/commit/6001dc05a9ae6aaf47c32f4deb407bbb93e3fa61).

-Brian

PR #128, opened 2017-09-24
Updated Slovak language

stankolubomir:
* Number : Column header - (Name of the "Number" column header. Also used in the context menu to copy the number value to the clipboard.)
* String : Column header - (Name of the "String" column header. Also used in the context menu to copy the string value to the clipboard.)
* External plugins : Group header - (Name of the header for 3rd party plugins group.)                                          
* Built-in plugins : Group header - (Name of the header for the built-in plugins group.)


brianferguson:
Thanks!

-Brian

PR #129, opened 2017-09-25
Hungarian language updates

VeRychard:
Removed unused string for select language, and updated untranslated strings.


brianferguson:
Thank you for the updates. The `SELECTLANGUAGE` string is still used in the installer in some cases (even though it's an empty string). I will add it back.
[https://github.com/rainmeter/rainmeter/blob/master/Build/Installer/Installer.nsi#L138](https://github.com/rainmeter/rainmeter/blob/master/Build/Installer/Installer.nsi#L138)


-Brian

VeRychard:
Hi,

Ah alright, sorry for that. And thanks :)

Richard


On Sep 25, 2017 6:49 PM, "Brian" <notifications@github.com> wrote:

Thank you for the updates. The SELECTLANGUAGE string is still used in the
installer in some cases (even though it's an empty string). I will add it
back.
https://github.com/rainmeter/rainmeter/blob/master/Build/
Installer/Installer.nsi#L138

-Brian

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/rainmeter/rainmeter/pull/129#issuecomment-331942731>,
or mute the thread

PR #136, opened 2017-10-20
Update French.h

oXidFoX:
some corrections and additions.


oXidFoX:
"meters" are like objects ? that's way better than the french equivalent of "counter". And at least a french understandable word.

PR #144, opened 2017-12-25
Update PortugueseBrazil.h

rogeriolagares:
Translated : STR_LANGUAGEOBSOLETE


brianferguson:
Thank you for translations, however there are still 4 items that need translating.
[https://github.com/rainmeter/rainmeter/blob/master/Language/PortugueseBrazil.h#L95-L98](https://github.com/rainmeter/rainmeter/blob/master/Language/PortugueseBrazil.h#L95-L98)

-Brian

PR #147, opened 2018-01-25
UniqueRandom support for QuotePlugin

tjhrulz:
This PR makes a few changes to QuotePlugin to support the UniqueRandom option found in other parts of Rainmeter. It also makes a few tweaks to make this possible. 

To kind of summarize the changes:
Directory reading is basically unchanged we just move the vector indexes to another vector after they have been used and swap the vectors when all have been used.
File reading has pretty drastically changed however it is for the better and is 100% BWC. Before it would just chose a random char location find the next 2 separators and that would be the string. This meant that the first line had only a 1/charCount chance of being selected if our RNG was perfectly distributed. This also meant that separator pairs that were very spaced out would increase the likelihood of the next pair being selected. Now the separator count and locations are prechecked causing even odds and are checked on update to prevent pesky file changes and keep BWC. This then allows support for UniqueRandom to happen the same was as it is with directories.

Possible improvements:
Add directory watching so we can dynamically add or remove files instead of missing/loading files that no longer exist. (Not sure if we want to do this with how bad this stuff is in windows but it may be better when only looking for file add and remove)
DynamicVariables=1 should be discouraged slightly more than before but with some tweaks it may be possible to make them play nice.
Functions could be improved for better readability and closer code style.

### DIFF
--- a/Plugins/PluginQuote/Quote.cpp
+++ b/Plugins/PluginQuote/Quote.cpp
@@ -27,8 +27,10 @@ T GetRandomNumber(T size)
 struct MeasureData
 {
 	std::wstring pathname;
+	bool isUniqueRandom = false;
 	std::wstring separator;
 	std::vector<std::wstring> files;
+	std::vector<std::wstring> usedFiles;
 	std::wstring value;
 };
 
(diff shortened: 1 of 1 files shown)


tjhrulz:
DV=1 and SetOption are now well handled. The method is it will detect if an option that is important has changed (Although brian you had a concern about false positives due to changes in case) and if it has it will then redo the setup process for the options and starts over.

I also stopped just passing the Measure struct for all the functions. It still is a little messy but it was pretty hairy in places to begin with.

I still have not added file watching but for the amount of work that would take for very little gain I don't know if it is worth it.

PR #154, opened 2018-04-09
Added support for value type REG_BINARY for registry keys

chpock:


brianferguson:
I would think it would pretty difficult to "display" any binary data from the registry. Can you elaborate on this idea and/or give specific real world skin examples?

-Brian

chpock:
This patch gets the binary value of the registry key as a hexadecimal string. This value can be modified by other measures, the necessary bytes can be extracted and used to display any meters.

For example, I use this to get windows accent color from the registry. Discussion of the issue on the forum:
* https://forum.rainmeter.net/viewtopic.php?f=15&t=21303
* https://forum.rainmeter.net/viewtopic.php?t=21223

I got this idea from this skin:
* https://github.com/tjmarkham/win10widgets/blob/master/Win10%20Widgets/%40Resources/systemColors.ini

The solution is excellent. But leads to the storm of child processes of the command reg.exe, when used by many skins.

This patch is very simple and can not lead to any new errors. Just adds a small feature. It is possible that getting some values from the binary key of the register will also be useful for other measurements.

PR #155, opened 2018-04-15
Added support for value type REG_BINARY for registry keys

chpock:
Previous discussion: https://github.com/rainmeter/rainmeter/pull/154


poiru:
Thanks! Apologies for the long delay.

PR #157, opened 2018-04-21
AudioLevel Replace QUERY_TIMEOUT with GetNextPacketSize

alatsombath:
AudioLevel Replace QUERY_TIMEOUT with GetNextPacketSize

Instead of every ~16 milliseconds ([QUERY_TIMEOUT](https://github.com/rainmeter/rainmeter/blob/970ba451802abcdcc32b1f8babec11be0f3eed90/Plugins/PluginAudioLevel/PluginAudioLevel.cpp#L69) = 1/60 = 0.0166..7 seconds), get the buffer when [GetNextPacketSize](https://msdn.microsoft.com/en-us/library/windows/desktop/dd370860(v=vs.85).aspx) retrieves a value greater than zero

Behavior before 46a449c:

The buffer was retrieved every Update, as fast as 10 milliseconds based on [SetTimer](https://github.com/rainmeter/rainmeter/blob/46a449c8fc96c455d0d3884040aaa3a92cf8120f/Library/Skin.cpp#L423) ([USER_TIMER_MINIMUM](https://msdn.microsoft.com/en-us/library/windows/desktop/ms644906(v=vs.85).aspx) = 0x0000000A = 10)

This is because [this condition always evaluated to true:](https://github.com/rainmeter/rainmeter/blob/970ba451802abcdcc32b1f8babec11be0f3eed90/Plugins/PluginAudioLevel/PluginAudioLevel.cpp#L539) current quad part (positive whole number of ticks counted) grows larger than the polling (start) count, multiplied by [m_pcMult](https://github.com/rainmeter/rainmeter/blob/970ba451802abcdcc32b1f8babec11be0f3eed90/Plugins/PluginAudioLevel/PluginAudioLevel.cpp#L228) which is a fraction is a non-factor, so postive integer >= fraction always


alatsombath:
AudioLevel Conditionally calculate RMS/Peak values on Parent


Parent measure [returns RMS by default](https://github.com/rainmeter/rainmeter/blob/970ba451802abcdcc32b1f8babec11be0f3eed90/Plugins/PluginAudioLevel/PluginAudioLevel.cpp#L174). If measure Type is specified on Parent (or Child) as other than RMS or Peak, don't calculate them every Update for the Parent

PR #160, opened 2018-04-29
Added a list of meters in the about dialog

chpock:
The about dialog displays meters, their status (hidden or displayed),
coordinates and dimensions.

Example: https://i.imgur.com/cqsFB2p.png


brianferguson:
Thanks for the PR. We appreciate the effort and overall, we like the ability to display more skin information. However, we do not believe this would be the correct way to display this information given how the list box is geared mainly for measures and variables.

We will keep this in mind for future developments to the About dialog box.

Thank you.

-Brian

PR #172, opened 2018-06-02
Adds InterpolationMode option to Image meter. This option changes the…

FabioGNR:
… interpolation mode used for scaling.

My personal use case for this is sharper scaling when using ScaleMargins:
Before: 
image
After with HighQualityBilinear interpolation mode:
image

This might be useful for other users as well, hence this PR.
Please let me know if any changes are needed.

### DIFF
--- a/Library/MeterImage.cpp
+++ b/Library/MeterImage.cpp
@@ -123,6 +123,9 @@ void MeterImage::ReadOptions(ConfigParser& parser, const WCHAR* section)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


tjhrulz:
Also I talked to Azack about adding this as an option a few weeks back and likely it will happen once the move to D2D is made.

ckosmic:
Any updates on this?  I could really use this for a skin I'm making that upscales a small image but needs to retain its pixelated look.

PR #185, opened 2018-07-30
Fix potential hanging as a multithreaded app (ActionTimer)

alatsombath:
MeasurePlugin::UpdateValue() -> [System::ResetWorkingDirectory()](https://github.com/rainmeter/rainmeter/blob/47c9986fdd5f77078141e9c78c999fad79480042/Library/System.cpp#L1069-L1077) -> GetCurrentDirectory() / SetCurrentDirectory()

From [MSDN](https://docs.microsoft.com/en-us/windows/desktop/api/winbase/nf-winbase-getcurrentdirectory):

"Multithreaded applications and shared library code should not use the
GetCurrentDirectory function and should avoid using relative path names. The current directory state written by the SetCurrentDirectory function is stored as a global variable in each process, therefore multithreaded applications cannot reliably use this value without possible data corruption from other threads that may also be reading or setting this value. This limitation also applies to the SetCurrentDirectory and GetFullPathName functions. The exception being when the application is guaranteed to be running in a single thread, for example parsing file names from the command line argument string in the main thread prior to creating any additional threads. Using relative path names in multithreaded applications or shared library code can yield unpredictable results and is not supported."

### DIFF
--- a/Library/MeasurePlugin.cpp
+++ b/Library/MeasurePlugin.cpp
@@ -16,11 +16,11 @@ MeasurePlugin::MeasurePlugin(Skin* skin, const WCHAR* name) : Measure(skin, name
 	m_ReloadFunc(),
 	m_ID(),
 	m_Update2(false),
-	m_MultiThreaded(false),
 	m_PluginData(),
 	m_UpdateFunc(),
 	m_GetStringFunc(),
-	m_ExecuteBangFunc()
+	m_ExecuteBangFunc(),
+	m_OverrideDirectoryFunc()
 {
 }
 
@@ -70,7 +70,7 @@ void MeasurePlugin::UpdateValue()
 		}
 
 		// Reset to default
-		if (!m_MultiThreaded)
+		if (!m_OverrideDirectoryFunc)
 		{
 			System::ResetWorkingDirectory();
 		}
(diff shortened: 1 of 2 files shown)


alatsombath:
Also commenting out this block of debug code appears to help with other skins lagging: https://github.com/rainmeter/rainmeter/blob/10a509b6277a07599031c57294b9fcfb9e25e409/Common/Gfx/Canvas.cpp#L251-L261

brianferguson:
> therefore multithreaded applications cannot reliably use this value without possible data corruption from other threads that may also be **reading** or **setting** this value.

ActionTimer isn't calling Get/SetCurrentDirectory, so I am not sure there is any issue here. Can you provide an example that is not working?


> Also commenting out this block of debug code appears to help with other skins lagging:

This isn't part of any debug code, and is used in connection with determining if a pixel is transparent (for mouse detection). However, this is part of the 4.3 alpha testing we are currently doing and is subject to change.

-Brian

alatsombath:
As you mentioned correctly it does not appear to be a multithreading issue. However when Rainmeter is installed on my flash drive, it gets stuck in that ResetWorkingDirectory function and the program becomes unresponsive, because it seems ActionTimer is calling to update (another Plugin) too quickly (IgnoreWarnings was set to 0)

If the issue is hardware-dependent and where Rainmeter is located, then Wait times cannot be reliably used below an undetermined inconsistent threshold (~10ms on my USB device). It reduces the effectiveness of ActionTimer for non-interactive Meters that are using data from Plugins

I added an "OverrideDirectory" function which can be used by Plugins that do not foresee changing the working directory of the application. If present, it allows Rainmeter to bypass calling Get / SetCurrentDirectory functions on every update of that Plugin

For 4.3 it appears to be better performing now, good work on the fixes for the EndDraw/CopyFromBitmap issue and improvements on that front

alatsombath:
See https://github.com/alatsombath/Fountain-of-Colors/releases/tag/v3.8.2

PR #190, opened 2018-11-20
Updated Swedish.h

tobzion:
Added some missing Strings!


brianferguson:
Could you also update the `STR_LANGUAGEOBSOLETE` string as well?
[https://github.com/rainmeter/rainmeter/blob/master/Language/Swedish.h#L11](https://github.com/rainmeter/rainmeter/blob/master/Language/Swedish.h#L11)

-Brian

PR #194, opened 2018-12-11
Allow images, that are open for writing, be loaded.

d-uzlov:
This is primarily allowing plugins to create images without need to write anything to disk.

Construction like this
`CreateFileW(path, GENERIC_WRITE, FILE_SHARE_READ | FILE_SHARE_DELETE, nullptr, CREATE_ALWAYS, FILE_ATTRIBUTE_TEMPORARY, nullptr);`
allows you to have a file in memory. But only until the handle is open.

r3266 can't open such files because it doesn't specify FILE_SHARE_WRITE mode, and handle with write access can't be closed as this will cause flush file data to disk.
This commit fixes this.

### DIFF
--- a/Common/Gfx/Util/D2DBitmapLoader.cpp
+++ b/Common/Gfx/Util/D2DBitmapLoader.cpp
@@ -22,7 +22,7 @@ HRESULT D2DBitmapLoader::LoadBitmapFromFile(const Canvas& canvas, D2DBitmap* bit
 
 	HANDLE fileHandle = CreateFile(
 		path.c_str(),
-		GENERIC_READ, FILE_SHARE_READ,
+		GENERIC_READ, FILE_SHARE_READ | FILE_SHARE_WRITE,
 		nullptr,
 		OPEN_EXISTING,
 		FILE_ATTRIBUTE_NORMAL | FILE_FLAG_SEQUENTIAL_SCAN,
@@ -127,7 +127,7 @@ bool D2DBitmapLoader::HasFileChanged(D2DBitmap* bitmap, const std::wstring& file
 
 	HANDLE fileHandle = CreateFile(
 		file.c_str(),
-		GENERIC_READ, FILE_SHARE_READ,
+		GENERIC_READ, FILE_SHARE_READ | FILE_SHARE_WRITE,
 		nullptr,
 		OPEN_EXISTING,
 		FILE_ATTRIBUTE_NORMAL | FILE_FLAG_SEQUENTIAL_SCAN,
@@ -154,7 +154,7 @@ HRESULT D2DBitmapLoader::GetFileInfo(const std::wstring& path, FileInfo* fileInf
 
 	HANDLE fileHandle = CreateFile(
 		path.c_str(),
-		GENERIC_READ, FILE_SHARE_READ,
+		GENERIC_READ, FILE_SHARE_READ | FILE_SHARE_WRITE,
 		nullptr,
 		OPEN_EXISTING,
 		FILE_ATTRIBUTE_NORMAL | FILE_FLAG_SEQUENTIAL_SCAN,


brianferguson:
A couple of questions:

1. What happens if Rainmeter is trying to read the file, and another program (or a threaded plugin) is currently writing to it? Feels like some collision would happen which could result in missing or corrupt data.

2. According to [this](https://docs.microsoft.com/en-us/windows/desktop/api/fileapi/nf-fileapi-createfilea#parameters) (under `dwShareMode`) Microsoft document, "You cannot request a sharing mode that conflicts with the access mode that is specified in an existing request that has an open handle." Note: It doesn't specify what constitutes a "conflict".

    In your example, you are using sharing modes Read and Delete (and not Write), does that mean Read and Delete also need to specified on the Rainmeter side as well?

    I guess this comes down to the fundamental question of, since Rainmeter does not attempt to write or delete anything, isn't FILE_SHARE_READ enough? Wouldn't that still be compatible with another "open" file handle?

    According to the table on [this](https://docs.microsoft.com/en-us/windows/desktop/fileio/creating-and-opening-files#creating-and-opening-files) Microsoft document, FILE_SHARE_READ should suffice, although it does not contain FILE_SHARE_DELETE in those examples.

3. Normally I like little changes like this that have the potential to open the doors to more creative ways to use Rainmeter. However, in this case, any benefit this brings feels narrow in scope and likely to only be used to solve a very narrow set of problems (if any).

    Can you give us a variety of different types of problems that this will solve?

-Brian

d-uzlov:
1. If the writing happens in the exact same time when Rainmeter reads the file then read data could potentially be corrupted.
However, if the file is written from Rainmeter main thread (like from plugin Update function), everything will be okay.
In case of someone writing the file in one piece there should also be no problems.
The worst that could happen if file is corrupted is that Rainmeter will show not consistent image—in the rare case when corrupted data looks valid. But the most common behavior would be to ignore the file altogether, just like now (when file can not be opened at all).
And reading the file that is being written is not the only way to get corrupted data. If the program, that writes the file, do something wrong, or crashes, the file becomes corrupted. Rainmeter already ignores such files.

2. dwShareMode is prohibitive. If you don't specialize X to do with file, then no one else can open the file to do X. However, if you specialize X, it doesn't mean that other programs should do X.
And if someone has already opened the file, then dwShareMode should at least specialize all the rights with which file is opened.
If you specialize Read, Write in dwShareMode when opening the file, it merely means that anyone can specify dwDesiredAccess to Read and/or Write.
If you specialize only Read in dwShareMode, other programs can read file, but can not write. Or, if someone has already opened the file for writing, you can't open the file.

If someone else opens the file with Read access, the share mode should include Read. If someone else opens the file with Write access, then share mode should include Write.
If only one program opens the file, then share mode do nothing.
If two programs open the file, then Program1 should set dwShareMode to the values, that Program2 set in dwDesiredAccess, and reverse.
For example:
Program1: dwDesiredAccess=Read, dwShareMode=Write
Program2: dwDesiredAccess=Write, dwShareMode=Read
Both Program1 and Program2 can add anything else into dwShareMode, nothing will change.

FILE_SHARE_DELETE is special right. Read and Write apply to subsequent calls to CreateFile, Delete does not. FILE_SHARE_DELETE should have probably be named FILE_SHARE_RENAME, but Microsoft liked word "delete".
It allows you to rename file, like in any UNIX. Deleting is actually just renaming to something inaccessible.
_Not_ specializing FILE_SHARE_DELETE is like locking the file so no one can rename it. I added Delete share mode here because I copied it from my program, where I want the file to be renameable, not because it's critical for opening files.

3. I see two general cases where it could be applied.
3.1. Generating images in memory using some data. You can define any custom Measure.
You could create some modified meter. For example, draw some text with subpixel antialiasing. Or Line that doesn't move, and instead overwrites old values (like ring buffer).
And, for example, Spectrogram, that I mentioned in my post in Rainmeter Forum, can not be created with any of the existing meters. And, I think, it's won't ever be possible, because such meter would require to transfer about 100 parameters from other meters, and it's unlikely that someone will want to write such skin.
3.2. Obtaining data from somewhere and showing it in Rainmeter.
This includes animation, downloading image from the internet or lan, reading images from devices (web-camera, etc.), exctracting images from files (for example, from tags of .mp3 file), creating images from unsupported image formats.

If something changes, for example, every minute, then keeping it in the cache, instead of writing to disk, can already benefit you. And if something changes every second, or even every frame, then the cache is just essential.

d-uzlov:
It worth noting, that files, that are not images, and image files, that weren't properly written aren't different.
So if providing arbitrary file path to Image meter is allowed, then providing path to potentially incorrect image will change absolutely nothing.

PR #195, opened 2019-01-19
Compatibility with the Nahimic audio driver

SnGmng:
The AudioLevel plugin doesnt work on PCs that use the Nahimic audio driver, this commit fixes this issue.
The issue is caused by Nahimic (the audio driver of MSI) because it is hooked into GetMixFormat and returns a wrong WAVEFORMATEX 
see: https://social.msdn.microsoft.com/Forums/windowsdesktop/en-US/bd8cd9f2-974f-4a9f-8e9c-e83001819942/iaudioclient-initialize-failure


d-uzlov:
This will also completely disable support for 7.1 format.

SnGmng:
> This will also completely disable support for 7.1 format.

Oh i didnt think of that. Anyways i fixed it in my last commit by changing the WAVEFORMATEX only if the audioclient initialization failed

PR #203, opened 2019-08-05
Meter support for Plugins

SnGmng:
This PR will allow future Plugins to add new Meters and Measures instead of only new Measures.

- Plugins will be able to add custom meters.
- The functions to add a new Meter are completely optional.
- The PR doesnt make any breaking changes to the plugin API whatsoever.
- Plugin Meters can bind with any Measure available.
- If a Plugin Meter binds with a Plugin Measure, it gets access to the pluginData of the measure too.

The Plugin will be able to use following new additional functions:

- MeterInitialize
    - Initialize your Meter here
    - Parameters:
        - `void** pluginData`: Same purpose as in normal Initialize().
        - `void* rm`: Same purpose as in normal Initialize().
- MeterFinalize
    - Finalize your Meter here
    - Parameters:
        - `void* pluginData`: Same purpose as in normal Finalize().
- MeterReload
    - Reload meter options here
    - Parameters:
        - `void* pluginData`: Same purpose as in normal Reload().
        - `void* rm`: Same purpose as in normal Reload().
- MeterUpdate
    - Update the meter here
    - Parameters:
        - `void* pluginData`: The pluginData that was initialized in MeterInitialize().
        - `void* pluginMeasureData`: If the measure that is bound to the meter is from the same plugin, the pluginData that was initialized in the Measure Initialize() is supplied too.
        - `double value`: The double value from the bound measure.
        - `LPCWSTR string`: The string value from the bound measure.
- MeterDraw
    - Draw the meter here
    - Parameters:
        - `void* pluginData`: The pluginData that was initialized in MeterInitialize().
        - `void* pluginMeasureData`: If the measure that is bound to the meter is from the same plugin, the pluginData that was initialized in the Measure Initialize() is supplied too.
        - `ID2D1DeviceContext* target`: The target to draw the measure on.
    - The ID2D1DeviceContext is handled from Gfx::Canvas. Passing a Gfx::Canvas ponter would be better but is not possible because Gfx::Canvas is unable to be compiled standalone (has too many dependencies in Library)

Here is a small example of the functions:

Screenshot_86

The usage of MeterPlugin is quite similar to MeasurePlugin. Here is an example:
Screenshot_87

I hope you like this idea, have a good day! 👍


SnGmng:
Hey its been half a year since i opened this PR. I just wanted to ask whats the current situation? Does this feature have to be delayed until some major milestone is archieved? I am just wondering.
Is everyone agreed about the current api spec?

I really need this PR because it gives me the power to create amazing skins,
and this PR will allow many plugin developers to do so many things such as:

- draw complex shapes much more efficient 
- render 3d objects (if we add `ID3D1DeviceContext` to the spec)
- use advanced effects such as blur
- improve performance by offloading heavy work to the plugin

I would be very happy to hear you guys opinion about this! 👍

cherryleafroad:
Offloading drawing to the plugin would be amazing for saving CPU and increasing performance. Lua visualizers are wayyy too CPU intensive.

brianferguson:
Thank you for the PR. The team has discussed meter plugins/scripts various times over years, but we haven't yet pursued a course of action.

Each time we discuss methods similar to this, there is one glarring issue that keeps coming up that prevents us from accepting this approach. Backwards compatibility is a major priority for us. This means we cannot expose any rendering specific "context" to 3rd parties, since this would lock us down to the current rendering system forever. Even minor changes or upgrades to different D2D rendering methods could break any meter plugin at any time. Over the years, we have used GDI, GDI+, a combination of GDI+/D2D, and now we mostly use D2D to render items. Who knows what we will use in the future.

This doesn't mean *no* to meter plugins in general, but there would have to be some sort of API to the Rainmeter basic drawing functions (eg. DrawLine). This way if the rendering system changes, we *should* be able to adjust any options to the new system.

___

It has been suggested a few times that instead of a meter plugin interface, a script meter `Meter=Script` with a Lua API might be a better approach. This would give users the ability of editing the script to their needs without needing to recompile anything. Like above, it would require some calls to our drawing functions like DrawLine, etc.

___

We are also open to new meters as well.

-Brian

PR #206, opened 2019-08-29
i18n, i10n: make 1 string translatable and improve Chinese (traditional) translations.

pan93412:


pan93412:
@brianferguson

brianferguson:
Thanks for the PR. I will update your translations for the first 4.4 beta.

Regarding the "Use hardware acceleration" string...there are still a few more steps to make that string translatable for all languages. We also may be adding some more strings in the near future, so we are going to wait until then to translate this string.

I am going to leave this PR open until then.

-Brian

PR #219, opened 2020-02-04
Update PortuguesePortugal.h

smarquespt:


Blackspirits:
I agree with only half the change. Others do not make sense and we are also translating into Portuguese, and you have modified some to English as is the case with Backup ...

smarquespt:
@Blackspirits Thanks for your comments. if you have write acesss please commit as you've said.

poiru:
Closing this in favor of #229.

PR #225, opened 2020-04-05
Fixed autoscaling for Line and Histogram meters and added additional autoscaling functionality

iranrmrf:
The MinValue and MaxValue for both meters now linearly scales between the two.

AutoScale for the Line meter can be now be dependent on a measure by using AutoScaleMeasure=#index#. AutoScale=1 is not needed if an AutoScaleMeasure index is supplied but nonetheless it still has its full functionality.

AutoScale for a Histogram meter using two measures can now scale each measure independantly, on primary, on secondary or on the maximum of the two (primary and secondary somewhat redundant most of the time). The setting is AutoScaleType and valid inputs are: "Independent", "Primary", "Secondary", "Maximum". The AutoScale option still works similarly as in the Line meter case and defaults to independent autoscaling. AutoScaleType can be ommited when AutoScale=1 is used.


poiru:
I'm not sure if anyone saw this cc @brianferguson

brianferguson:
Thanks for the reminder. I haven't had time to fully review and investigate any backward compatibility issues it may cause. This is on my ever growing "todo" list. :dizzy_face:

-Brian

PR #230, opened 2020-05-28
Update PingPlugin to handle both IPv4 and IPv6 addresses

reidrankin:
This is an update/rebase of #110 on a feature-branch. It also eliminates the deprecation warning overrides added in c9b5118, because this patch no longer uses deprecated functions.


brianferguson:
Thanks for this PR, we really appreciate it!

-Brian

PS - Sorry it took so long to review and merge.

PR #231, opened 2020-05-29
Fix a bug with the !EnableMeasure bang that silently fails if Dynamic…

DoubleStrike:
…Variables are turned off

This is a bug I found when trying to troubleshoot why the `!EnableMeasure` bang was not working consistently when called from a WebParser measure's `FinishAction`.  Please review for inclusion into master.

### DIFF
--- a/Library/Skin.cpp
+++ b/Library/Skin.cpp
@@ -1490,7 +1490,10 @@ void Skin::EnableMeasure(const std::wstring& name, bool group)
 	{
 		if (CompareName((*i), measure, group))
 		{
+			// The Enable() call will fail unless DynamicVariables are forced on
+			(*i)->SetDynamicVariables(true);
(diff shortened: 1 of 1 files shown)


brianferguson:
Sorry it took so long to get to this.

I am not sure what this is solving, could you provide an example where `!EnableMeasure` does not work?

This works:
``` ini
[Test_1]
Measure=Calc
Formula=Test_1+1
Disabled=1

[MeasureIPWeb]
Measure=WebParser
Flags=Resync | NoCookies
URL=https://checkip.amazonaws.com/
RegExp=(?siU)^(.*)$
StringIndex=1
UpdateRate=30
RegExpSubstitute=1
Substitute="^$":"None"
FinishAction=[!EnableMeasure Test_1]
```

-Brian

PR #240, opened 2020-10-04
PingPlugin enhancements

### LINKED ISSUE #239: PingPlugin: Spurious RTT values for IPv6 destinations on systems without IPv6 support
Systems without IPv6 support will sometimes give strange (usually very high) readings for ping time to IPv6 addresses.

This is caused by failing to check the `IcmpSendEcho2`/`Icmp6SendEcho2` return value. A zero return value indicates an error, and that the (uninitialized) reply buffer has not been changed. Occasionally, the initial contents of the uninitialized buffer will contain an IP_SUCCESS (zero) value in the `Status` field; this is treated as a success, and results in the garbage `RoundTripTime` field being reported as a correct value.

reidrankin:
Fixes #239 by properly checking the return values of `IcmpSendEcho2`/`Icmp6SendEcho2` and zero-initializing the reply buffer. Also adds descriptive error messages when reporting failures.

### DIFF
--- a/Plugins/PluginPing/Ping.cpp
+++ b/Plugins/PluginPing/Ping.cpp
@@ -53,7 +53,7 @@ struct MeasureData
 	{
 		if (destAddrInfo)
 		{
-			FreeAddrInfo(destAddrInfo);
+			FreeAddrInfoW(destAddrInfo);
 			destAddrInfo = nullptr;
 		}
 	}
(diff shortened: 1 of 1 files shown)


reidrankin:
(I changed to GetAddrInfoW/FreeAddrInfoW before fixing the issue, but I rebased so it's at the end so it's easier to pull just the error-handling bit if you'd like.)

poiru:
Thank you!

PR #264, opened 2021-06-21
Update ChineseSimplified.h

Lssg97:
Hi @brianferguson,

The reason I initiated this PR is that I noticed that you changed the Chinese quotation marks in the translated text to English quotation marks in 66b274d50242a135ecfe775d0debc80652077a81. Although I don't know why English quotation marks are used, in written Chinese, a **space** need to be added at the alternation of Chinese and English. The same is true for punctuation, otherwise it will be visually crowded (As shown below). So I added it by referring to the old translated text format.

image


brianferguson:
Ah, I see. Thanks for the clarification. Often times user submissions can contain small typos or omissions. I saw the Chinese quotation marks and changed them to the English version to be consistent with the rest of the translations. Like these: [https://github.com/rainmeter/rainmeter/blob/master/Language/ChineseSimplified.h#L181-L210](https://github.com/rainmeter/rainmeter/blob/master/Language/ChineseSimplified.h#L181-L210)

I failed to see that the Chinese quotation marks contained extra space for readability. Thank you for checking!

-Brian

PR #265, opened 2021-06-23
Corrected (translated from English) and updated Ukrainian translation.

YuriPet:


brianferguson:
Thank you for the translations, however, escaped quotation marks `\"` **cannot** be accepted.

Please use a single quotation mark in the .nsh file: `'`
And use double quotes in the .h file: `""` (2 quotation marks, instead of an escaped quotation mark).

Also, please do not add extra quotation marks in places that do not have those quotation marks in the English files.

Here are what the English files look like:
[https://github.com/rainmeter/rainmeter/blob/master/Language/English.h](https://github.com/rainmeter/rainmeter/blob/master/Language/English.h)
[https://github.com/rainmeter/rainmeter/blob/master/Language/English.nsh](https://github.com/rainmeter/rainmeter/blob/master/Language/English.nsh)

Please re-submit your pull request once completed.

Thank you,
-Brian

PR #270, opened 2021-08-21
Fix SysInfo Plugin GetBestInterfaceOrByName always return 0

aetasoul:
Reference on the [rainmeter forum](https://forum.rainmeter.net/viewtopic.php?f=5&t=38083)

Any adapter name used in `SysInfoData=` always returns 0, except for `SysInfoData=BEST` that return the correct value.

### DIFF
--- a/Plugins/PluginSysInfo/SysInfo.cpp
+++ b/Plugins/PluginSysInfo/SysInfo.cpp
@@ -19,6 +19,8 @@
 #include "../../Common/StringUtil.h"
 
 #define INADDR_ANY (ULONG)0x00000000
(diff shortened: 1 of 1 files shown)


brianferguson:
@aetasoul 
I believe we have fixed this issue. Could you try out the following test build and report back if it works for you?
[https://forum.rainmeter.net/viewtopic.php?f=143&t=38371#p197283](https://forum.rainmeter.net/viewtopic.php?f=143&t=38371#p197283)

Thank you,
-Brian

aetasoul:
Hi @brianferguson 

I'll test tomorrow and let you know! Thank you for your work!

aetasoul:
Hi @brianferguson sorry for the dilay but I was out of office. All is working correctly

PR #271, opened 2021-08-28
Translating game mode to Polish

Adamo2499:
Translating things about game mode to Polish


brianferguson:
Looks like there is still 1 more item still to be translated:
[https://github.com/rainmeter/rainmeter/blob/master/Language/Polish.h#L143](https://github.com/rainmeter/rainmeter/blob/master/Language/Polish.h#L143)

-Brian

PR #273, opened 2021-09-06
Fix NowPlaying for Spotify Premium by making it scan the windowTitle for 'Spotify'

FastFerrari5555:
By default, NowPlaying in Rainmeter looks at the length of the window title, and makes sure it is longer than 7 characters (the length of "Spotify"), in order to detect playing media. When using Spotify Premium, it is default longer than 7 characters, so it thinks a song is always playing, and does not return the correct value for the state of the player. By scanning for the string "Spotify" in the window title, you can determine whether the media is paused, regardless of whether you use free or premium

### DIFF
--- a/Library/NowPlaying/PlayerSpotify.cpp
+++ b/Library/NowPlaying/PlayerSpotify.cpp
@@ -103,11 +103,31 @@ void PlayerSpotify::UpdateData()
 {
 	if (m_Initialized || CheckWindow())
 	{
-		// Parse title and artist from window title
 		WCHAR buffer[256];
 
-		//Length of window is now 7 when not playing
-		if (GetWindowText(m_Window, buffer, 256) > 7)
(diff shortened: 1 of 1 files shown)


FastFerrari5555 (commented):
I forgot to add, I built my own version of Rainmeter with this code. It uses a bunch of if statements, which I realize is maybe a little crude, but it works just fine, and I'm happy with my new skin I created with it.

brianferguson:
I uploaded a fix for this, could you test this out and report back if it works for you?
[https://forum.rainmeter.net/viewtopic.php?f=143&t=38371](https://forum.rainmeter.net/viewtopic.php?f=143&t=38371)

Thank you,
-Brian

FastFerrari5555:
It works just fine!  A good function like _wcsnicmp was what I was trying to use, but I was looking at the other variants of that, and couldn't figure it out, since all I know is C#. Glad you converted my crude code to something good!

PR #280, opened 2021-10-30
Update PlayerWinamp.cpp

dr0biwan:
The return buffer of IPC_GET_PLAYING_FILENAME & other Winamp api that return a string aren't guaranteed to always be at the same memory location in the running instance.

This change accounts for that plus it also makes things work better with WACUP (an effort to make a Winamp 5.666 compatible replacement) which would only show the first item played with how the existing code worked.


brianferguson:
Thank you!

-Brian

PR #283, opened 2021-11-11
Added OpenPlayer support for AIMP4 and up

deathcrafter:
- Check `HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall\` if system is 64-bit, more reliable.

- Also altered the order of checking executable path for AIMP, giving priority to latest versions.

- Tested and it works with all versions, from 2 up to 5.


deathcrafter:
Ahh, and yes, I couldn't test it on 32-bit system. May be do a pre-release to test on 32-bit systems, if you decide pull it in or make the changes.

brianferguson:
Thanks!

-Brian

PR #285, opened 2021-11-16
Update SerbianLatin.h

netizenk:
Completed Serbian Latin localization file.
Completed Serbian Cyrillic localization file.


brianferguson:
Thanks!

-Brian

PR #286, opened 2021-11-30
Fix: Range-check volume value for CAD plugin

leezer3:
The NowPlaying measure expects volume values to be within the range of 0- 100.

However, when recieving a volume value, the CAD plugin fails to range-check it, and just drops it in regardless, which can cause unexpected results.

I noticed this using JRiver Media Center 28 to playback images, where it's CAD interface seems to return uint.MaxValue (as images have no volume) for the volume.
Whilst this is probably a JRiver bug at the end of the day, it seems sensible to range-check the value is what we expect before passing it to the measure, as large values can mess up skins in interesting ways.


brianferguson:
Thanks!

-Brian

PR #300, opened 2022-04-24
add initial support for clang

NearTox:
This change can lead to improved execution speed and improved code robustness, reporting bugs for code misbehavior (see currently disabled bugs in Build\VS\Rainmeter.Cpp.props:54 for posible review and hardening)

the project is currently only configured for work with x64 builds for clang and I disabled tests(due a missing library)


brianferguson:
Please try to keep each pull request related to 1 specific topic.

I don't think we need clang support at this time, unless you can provide some evidence of "improved execution speed". If there is significant improvement in speed, please make another PR with just the specific changes for clang support.

Thanks,
Brian

PR #309, opened 2022-08-25
Use latest ilasm.exe available in DLLExporter.exe

deathcrafter:
The previous version used `TargetDotNetFrameworkVersion.Version20` to find ilasm.exe which was unable to assemble latest C# libraries.

### DIFF
--- a/Build/DllExporter/Program.cs
+++ b/Build/DllExporter/Program.cs
@@ -1,4 +1,4 @@
-﻿/* Copyright (C) 2011 Rainmeter Project Developers
+/* Copyright (C) 2011 Rainmeter Project Developers
  *
  * This Source Code Form is subject to the terms of the GNU General Public
  * License; either version 2 of the License, or (at your option) any later
@@ -7,16 +7,15 @@
 
 using System;
 using System.Collections.Generic;
-using System.Text;
 using System.Diagnostics;
-using Microsoft.Build.Utilities;
 using System.IO;
+using Microsoft.Build.Utilities;
 
 namespace DllExporter
 {
-    class Program
+    public class Program
     {
-        static int Main(string[] args)
+        public static int Main(string[] args)
         {
             if (args.Length < 4)
             {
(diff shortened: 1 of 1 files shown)


brianferguson:
Yeah, VS2022 doesn't play nice with `.NET 2.0`, which makes sense. We did update the `DllExporter.csproj` to use `.NET 4.7.1`.
[https://github.com/rainmeter/rainmeter/commit/013a2943895a846dd1e2e51e76d9f55b2e247793?diff=unified#diff-181d9da96ab0b054a87eb9c6ec63ce39f3602df43f82c0243178b198b918846f](https://github.com/rainmeter/rainmeter/commit/013a2943895a846dd1e2e51e76d9f55b2e247793?diff=unified#diff-181d9da96ab0b054a87eb9c6ec63ce39f3602df43f82c0243178b198b918846f)

However, I missed this part concerning `TargetDotNetFrameworkVersion`.

Correct me I am wrong, but I think we should not use `.VersionLatest` since we cannot guarantee that the end-user will have the correct version installed on their machine? 

For instance, if a plugin is built with `.NET 5.0`, but the end-user's machine only has `.NET 4.8` installed, won't the plugin fail?

What about [.Version471](https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.utilities.targetdotnetframeworkversion?view=msbuild-17#microsoft-build-utilities-targetdotnetframeworkversion-version471) ?

We chose to update to `.NET 4.7.1` since that ships with Windows 7, so everyone _should_ have it.

#

Also, it's good practice to separate different "issues" in separate PR's (or commits). So, I am going to close this PR since it includes all those "string to var" edits. Please re-submit this PR with just the changes to the `.Version20` to `.Version471` (if possible).

Thanks!

-Brian

Note: The build for this PR failed due to reasons unrelated to your PR. I am fixing them currently, so don't worry about them.

deathcrafter:
VersionLatest will use ilasm.exe from the latest `C:\Windows\Microsoft.NET\Framework64\ilasm.exe` or the `C:\Windows\Microsoft.NET\Framework\ilasm.exe`, depending on the architecture, which is installed on your device if you have `.NET Framework 4.0+`, which covers `4.7.1`.

And the plugin SDK uses `.NET Framework` which is completely different from `.NET`. So, it should be fine.

About the vars, just that I don't like wavy lines in my code LoL. I'll just open a new PR with only with the mentioned change.

PR #312, opened 2022-09-03
Enable Lua Update function to return two-element tables

deathcrafter:
This enables the Update function in the script measure to return a two-element table of which the first value will be the number value, and the second, the string value of the measure.

Example:
```lua
function Update()
    return { 24, "TwentyFour" }
end
```

Usage:
```ini
[ResultText]
Meter=String
MeterStyle=GithubCodeBlockStyle
Text="Num = [MeasureScript:] | Str = [MeasureScript]"
```
Result:
```txt
Num = 24 | Str = TwentyFour
```

### DIFF
--- a/Library/MeasureScript.cpp
+++ b/Library/MeasureScript.cpp
@@ -69,7 +69,7 @@ void MeasureScript::UpdateValue()
 */
 const WCHAR* MeasureScript::GetStringValue()
 {
-	return (m_ValueType == LUA_TSTRING) ? CheckSubstitute(m_StringValue.c_str()) : nullptr;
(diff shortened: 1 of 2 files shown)


brianferguson:
Now that I look through the code, I think we simplify this a bit to accept 2 return values, instead of parsing through a table.

Example:
``` lua
function Update()
  return 24, "Twenty four"
end
```

Would this be acceptable?

-Brian

deathcrafter:
Ohh how could I not think about this lol. I'll make the changes and reopen the pr. Thanks for this 😁

brianferguson:
No need, I have the code working.

-Brian

deathcrafter:
Ohh that's a far more elegant solution 😁. Gotta work on my logics.

PR #321, opened 2022-09-30
Fix skin draggable checkbox not updating state when global setting is changed

Joehuu:
| Before | After |
| --- | --- |
| Rainmeter_730tVPbf5a | Rainmeter_EkLYOnj077 |


brianferguson:
Thanks!

-Brian

PR #326, opened 2022-11-09
Fix blur not working in Windows10 or greater version

bhxxwx:
Fix blur not working in Windows10 or greater version (theoretically win8 or greater version, but win8 not tested), use SetWindowCompositionAttribute api. It worked fine in windows10 22H2. In windows10 use "Blur=3" command to force use Blur function.
Snipaste_2022-11-09_09-21-20


poiru:
Thanks for the PR! We plan to implement blur support using DirectComposition in the near future.

PR #341, opened 2023-04-09
WebParser: add functionality for JSON parsing

jschmer:
Hi, this PR adds functionality for JSON parsing to WebParser. Json matches are defined by using "json pointers" as explained below. Let me know if that is a valuable addition and I'll do my best to bring this pull request in shape according to your guidance.

JSON parsing is enabled by specifying
  ParseMode=Json

Once JSON parsing is enabled, the configurations in
JsonValuePath<N> are specifying the JSON pointer expressions
that should be applied to the JSON data. <N> is >= 1.
JSON pointers are explained in
https://github.com/nlohmann/json#json-pointer-and-json-patch

Example:
  [JsonParser]
  Measure=WebParser
  Url=file://#CURRENTPATH#weather-api.json
  ParseMode=Json
  JsonValuePath1=/id/days/1/temperatureMin
  JsonValuePath2=/id/days/2/temperatureMin
  JsonValuePath3=/id/days/3/temperatureMin
  JsonValuePath4=/id/days/4/temperatureMin
  JsonValuePath5=/id/days/4/doesNotExist
  JsonValuePath6=/id/days/1

In this case, the file weather-api.json will be parsed as
a JSON file and the expressions in JsonValuePath1 to
JsonValuePath6 are applied. The matches of these expressions
can be referenced by other measures by the usual StringIndex
mechanic. StringIndex=0 references the whole JSON data, similar
to Regex mode, e.g. this measure will have the whole JSON file
as its value.
JsonValuePath5 in the example above references data that does
not exist, e.g. the evaluation will fail. JSON parsing will
assign the value "N/A" in this case for this expression.

Example for referencing WebParser with JSON parsing:
  [Day2TempMin]
  Measure=WebParser
  Url=[JsonParser]
  StringIndex=2

This measure will get the result of "/id/days/2/temperatureMin".

Regex mode has the ability to reference the output of another
WebParser instance and apply another round of Regex. The same
applies to JSON parse mode.

Example:
  [JsonParserChild]
  Measure=WebParser
  Url=[JsonParser]
  ParseMode=Json
  StringIndex=6
  JsonValuePath1=/temperatureMin
  JsonValuePath2=/temperatureMax
  JsonValuePath3=/precipitation
  StringIndex2=3

This references the Index 6 output of [JsonParser], which is the
match for "/id/days/1". Then expressions JsonValuePath1 -
JsonValuePath3 are applied on that value and JsonValuePath3 is
taken for the value of this measure (as defined by StringIndex2=3).

### DIFF
--- a/.vscode/c_cpp_properties.json
+++ b/.vscode/c_cpp_properties.json
@@ -1,0 +1,17 @@
+{
+    "configurations": [
+        {
+            "name": "Win32",
+            "includePath": [
+                "${workspaceFolder}/**"
+            ],
+            "defines": [
+                "_DEBUG",
+                "UNICODE",
+                "_UNICODE"
+            ],
+            "compilerPath": "D:/Programme_x86/Microsoft Visual Studio/2022/Community/VC/Tools/MSVC/14.35.32215/bin/Hostx64/x64/cl.exe"
+        }
+    ],
+    "version": 4
+}
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,19 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "name": "Rainmeter",
+            "type": "cppvsdbg",
+            "request": "launch",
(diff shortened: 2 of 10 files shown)


poiru:
Thanks for the PR! We added JsonPointer support the latest [5.0 pre-release builds](https://forum.rainmeter.net/viewtopic.php?t=45991&start=10#p244174).

PR #346, opened 2023-05-09
Fix cursor changing to a hand when a meter has a mouse wheel action

Joehuu:
Not sure if intentional:

| Before | After |
| --- | --- |
| Rainmeter_enroN1Q62W | Rainmeter_Ght6XvBkIP |


brianferguson:
Thanks!

-Brian

PR #347, opened 2023-05-24
Fix skin meter `MouseLeaveAction` being delayed when skin is draggable and global dragging is disabled

Joehuu:
Here's a simple meter:
```ini
[MeterString]
Meter=String
Text=Unhovered
MouseOverAction=[!SetOption MeterString Text Hovered][!UpdateMeter MeterString][!Redraw]
MouseLeaveAction=[!SetOption MeterString Text Unhovered][!UpdateMeter MeterString][!Redraw]
SolidColor=255,255,255
```

Before:
hEU02neqe4

After:
i3lrYsndKZ

Note that I don't fully understand the codebase yet and the fix was found through trial and error.

### DIFF
--- a/Library/Skin.cpp
+++ b/Library/Skin.cpp
@@ -507,7 +507,7 @@ void Skin::SetMouseLeaveEvent(bool cancel)
 
 	// Set the mouse event
 	tme.dwFlags = TME_LEAVE;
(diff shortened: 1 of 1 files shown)


brianferguson:
Thanks!

-Brian

PR #348, opened 2023-05-24
Fix dialog global draggable checkbox not updating state when toggling via tray icon

Joehuu:
| Before | After |
| --- | --- |
| Rainmeter_dRUojb002i | Rainmeter_9dXXQZZryS |


brianferguson:
Thanks!

-Brian

PR #352, opened 2023-07-31
Fix: version comparison function

nopeless:
This crashed my build

### DIFF
--- a/Common/Version.h
+++ b/Common/Version.h
@@ -51,7 +51,7 @@ private:
 			int m = 0, n = 0;
 
 			while (i < size1 && m_Version[i] != L'.') { m = (m * 10) + (m_Version[i] - L'0'); ++i; }
-			while (j < size1 && version2[j] != L'.') { n = (n * 10) + (version2[j] - L'0'); ++j; }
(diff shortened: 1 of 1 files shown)


brianferguson:
Good catch.

I am curious though. That function is only called in 2 instances. 1 is the comparison between the "new" Rainmeter version versus the currently installed version. The other comparison is between the minimum Windows version (from the status file), and the current Windows OS version.

Do you happen to know which comparison crashed?

-Brian

brianferguson:
Good catch.

I am curious though. That function is only called in 2 instances. 1 is the comparison between the "new" Rainmeter version versus the currently installed version. The other comparison is between the minimum Windows version (from the status file), and the current Windows OS version.

Do you happen to know which comparison crashed?

-Brian

nopeless:
@brianferguson `0.0.0.0` against `4.5.17`

brianferguson:
Ok, I replicated this. Looks like it might only effect new installs, or at least ones that haven't done any update checking.

We will get a new version out right away.

Thanks!
-Brian

PR #353, opened 2023-07-31
Plugin input text refresh fix and features

nopeless:
Fixed some personally bugging things.

If something is needed, please comment

### DIFF
--- a/Plugins/PluginInputText/InputBox.cs
+++ b/Plugins/PluginInputText/InputBox.cs
@@ -20,7 +20,7 @@ namespace InputText
         private bool _Numeric = false;
         private string _TextValue = string.Empty;
 
-        private static Regex numberRegex = new Regex(@"^[-+]?(?>0b[10]*|0o[0-7]*|0x[A-Fa-f]*|\d*(?>\.\d*|[eE]\d*)?)", RegexOptions.Compiled);
+        private static Regex numberRegex = new Regex(@"^[-+]?(?>0b[10]*|0o[0-7]*|0x[A-Fa-f\d]*|\d*(?>\.\d*|[eE]\d*)?)", RegexOptions.Compiled);
 
         public InputBox(SkinWindow parent)
         {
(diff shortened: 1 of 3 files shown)


nopeless:
one sec

nopeless:
there we go

nopeless:
not sure what 
```
D:\a\rainmeter\rainmeter\x32-Release\Plugins\InputText.dll.il(488): error : syntax error at token '-' in:     IL_0062:  ldc.r8     -nan(ind) [D:\a\rainmeter\rainmeter\Plugins\PluginInputText\PluginInputText.csproj]
```
is

brianferguson:
We can't accept this PR for a few reasons.

1. The obvious one: the weird syntax error. Obviously something in the "Improve form validation" commit is causing it, but that would have to be resolved before we accept this.
2. Forcing an `!UpdateMeasure` is not a good idea. It can cause a premature firing of OnChangeAction and a double firing of OnUpdateAction. The former probably being okay, but the latter being not acceptable.
3. We prefer separate PR's for different items/topics. This helps with keep things organized, and  also helps determining "when" an issue (if any) came about.

-Brian

nopeless:
@brianferguson how would you fix 2. then? Refer me to resources/code i can use

Also, not sure what the syntax error is. The code compiles on my end. Do you have clues?

I'll split the PR

brianferguson:
> how would you fix 2. then? Refer me to resources/code i can use

Unfortunately the API doesn't have an "force the internal measure to update the value from the plugin" on demand type of function. A measure's value is only updated during the Update cycle or if the !UpdateMeasure bang is used. This is usually sufficient for most measures and plugins, but threaded plugins can get "out of date" until the next update cycle.

The best solution would be to have the skin author manually place a `!UpdateMeasure` bang in one of the `CommandN` options if they want a more instantaneous result without waiting for the next Update cycle.

> Also, not sure what the syntax error is. The code compiles on my end. Do you have clues?

Unfortunately, I have no idea either.

-Brian

PR #355, opened 2023-08-06
!ToggleConfig any args count

D-Maxwell:
The !ToggleConfig bang may now be called supplied only with a config, similarly to the !DeactivateConfig bang which it supposedly depends on.

Currently, calling !ToggleConfig without specifying a file will result in an error, despite the functionality being present in the two underlying bangs which it should rely on.

I have severely limited myself and my style, as to fit that of the rest of the code. Otherwise, I would have compacted this and made it a lot more modular.

In any case, I wish for this bug to be fixed.

Besides, please note that this has been an issue for 11 years, as stated [here](https://forum.rainmeter.net/viewtopic.php?p=216773#p216773).

### DIFF
--- a/Library/CommandHandler.cpp
+++ b/Library/CommandHandler.cpp
@@ -683,22 +683,26 @@ void CommandHandler::DoDeactivateSkinBang(std::vector<std::wstring>& args, Skin*
 
 void CommandHandler::DoToggleSkinBang(std::vector<std::wstring>& args, Skin* skin)
 {
-	if (args.size() >= 2)
+	if (skin)
 	{
-		Skin* skin = GetRainmeter().GetSkin(args[0]);
-		if (skin)
-		{
-			GetRainmeter().DeactivateSkin(skin, -1);
-			return;
-		}
-
-		// If the skin wasn't active, activate it.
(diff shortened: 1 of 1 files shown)


brianferguson:
Thanks for this PR, however, I cannot accept this for a couple of reasons.

1. The DeactivateSkin part doesn't work since the "sending" skin will be deactivated, not the skin defined by the parameter. Also, this bang will not longer work from outside of a skin, like from the command line.
2. There is not any checking of the ActivateSkin return value. This is need to determine if the parameters were valid.

I will be committing a fix for this very issue shortly.

-Brian

D-Maxwell:
Neat, thank you for your time, and please apologise my lack of understanding when it comes to Rainmeter as a whole.

Regardless, I am glad to see this being fixed.
May I get an ETA of the next Rainmeter update, so that I could make use of this ?

brianferguson:
I don't have an ETA on the next release. ☹️ 

If you would like to try out a prerelease version, you can find `Rainmeter-4.5.18.3730-prerelease.exe` test build [here](https://builds.rainmeter.net/test_builds/).

-Brian

PR #356, opened 2023-08-08
Refactor plugin input text

nopeless:
This meter will also return a number value if the number is a number format (returns NaN if it is not)

Number input validation is slightly better (also allow 0x, 0o, 0b prefix)

Clean up some code inconsistencies

I do not know if this resolves any issue

Build works on my machine

### DIFF
--- a/Plugins/PluginInputText/InputBox.cs
+++ b/Plugins/PluginInputText/InputBox.cs
@@ -7,6 +7,7 @@
 
 using System;
 using System.Drawing;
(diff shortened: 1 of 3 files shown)


poiru:
Thank you! InputText has been deprecated in Rainmeter 5.0 pre-release builds in favor of a [new meter](https://forum.rainmeter.net/viewtopic.php?t=46006) so we will not be accepting these changes.

PR #387, opened 2024-09-24
Repeat Feature for Measures and Meters

ryanlath:
This pr requests adds the ability to repeat measures and groups of meters.  The need for this should be obvious to any one who has tried to create any sort of list or table.  eg. Currently, 5 measures x 10 rows is 50 measures you have to create.  Yuck.  Add the 50 meters needed to display that and, well, no thanks!  This pr fixes that.

### Measures
Adds a "Reapeat=X" option where X is the number of times to repeat (starting at 1)
Measures will be created with the X suffix and 
#i# will be replaced in the measures options with X as well.

Example
```
; In your Meter.ini

[MeasureQuote]
Measure=WebParser
URL=#URLPrefix##Ticker#i##
Repeat=2
```
Would result in...
```
[MeasureQuote1]
Measure=WebParser
URL=#URLPrefix##Ticker1#

[MeasureQuote2]
Measure=WebParser
URL=#URLPrefix##Ticker2#
```

### Meters
Adds a Meter=Repeat that has two options
 - Count=X (the number of times for the group to repeat, starting at 1)
 - RepeatMeters=[comma separated list of meternames]
Like measures any #i# in Meter options will be replace with the Count value

Example:
```
; In your Meter.ini

[MeterQuoteName]
Meter=String
MeasureName=MeasureQuoteName#i#

[MeterQuotePrice]
Meter=String
MeasureName=MeasureQuotePrice#i#

[MeterRepeat]
Meter=Repeat
Count=2
RepeatMeters=MeterQuoteName,MeterQuotePrice
```
would result in
```
[MeterQuoteName1]
Meter=String
MeasureName=MeasureQuoteName1

[MeterQuotePrice1]
Meter=String
MeasureName=MeasureQuotePrice1

[MeterQuoteName2]
Meter=String
MeasureName=MeasureQuoteName2

[MeterQuotePrice2]
Meter=String
MeasureName=MeasureQuotePrice2
```

### Anyway
I've attached a Test.ini
A Test Meter demonstrating these features.
[Test.ini.txt](https://github.com/user-attachments/files/17106408/Test.ini.txt)

Output of test meter (the ugly square) and what I actually created this for a table of quotes that doesn't need 5k measures and meters to do it!
example2

### DIFF
--- a/Library/ConfigParser.cpp
+++ b/Library/ConfigParser.cpp
@@ -2213,7 +2213,7 @@ const std::wstring& ConfigParser::GetValue(const std::wstring& strSection, const
 ** Copies all values based on section name
 ** replace #i# with count value
 */
-void ConfigParser::CopySectionValuesWithRepeat(const std::wstring& fromSection, const std::wstring& toSection, uint32_t count)
+void ConfigParser::CopySectionValuesWithReplace(const std::wstring& fromSection, const std::wstring& toSection, uint32_t count)
 {
 	std::wstring strFrom;
 	strFrom.reserve(fromSection.size() + 1ULL);
--- a/Library/ConfigParser.cpp
+++ b/Library/ConfigParser.cpp
@@ -2265,29 +2265,28 @@ void ConfigParser::DeleteSectionValues(const std::wstring& section)
 ** Insert Section 
 ** 
 */
-void ConfigParser::SectionInsert(const std::wstring& fromSection, const std::wstring& toSection)
+void ConfigParser::InsertSection(const std::wstring& fromSection, const std::wstring& toSection)
 {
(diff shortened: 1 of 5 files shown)


chocmake:
This looks promising as an alternative to a Lua based approach with skin refreshes (I'm assuming the repeated meters would be virtual and not physically written to file?).

About the meter/measure naming: is different syntax possible? Eg: if one has meters that use the instance number in the middle of the meter name rather than at the end, such as `Foo_1_Bg`, `Foo_1_Str`, etc.

poiru:
Thanks for the PR! We have some ideas about how to implement this, but it will be using a different approach. Closing this!

PR #390, opened 2024-10-16
Python поддержка

SnayperTihCreator:
Можете добавить поддержку языка Python 3+


brianferguson:
Note: Please use english.

This PR doesn't seem to accomplish anything. It's just a python archive.

We won't be adding native python support to the script plugin (I am assuming that is what you are asking for).

Someone made a python plugin that can link your local python installation to Rainmeter.
[https://github.com/glitchassassin/rainmeter-python](https://github.com/glitchassassin/rainmeter-python)

-Brian

PR #396, opened 2025-03-02
Fix: Potential Vulnerability in Cloned Function

tabudz:
**Description**
This PR fixes a security vulnerability in f_parser() that was cloned from redis but did not receive the security patch. The original issue was reported and fixed under https://github.com/redis/redis/commit/fdf9d455098f54f7666c702ae464e6ea21e25411.
This PR applies the same patch to eliminate the vulnerability.

**References**
https://nvd.nist.gov/vuln/detail/CVE-2015-4335
https://github.com/redis/redis/commit/fdf9d455098f54f7666c702ae464e6ea21e25411

### DIFF
--- a/Library/lua/lua/ldo.c
+++ b/Library/lua/lua/ldo.c
@@ -495,7 +495,7 @@ static void f_parser (lua_State *L, void *ud) {
   struct SParser *p = cast(struct SParser *, ud);
   int c = luaZ_lookahead(p->z);
   luaC_checkGC(L);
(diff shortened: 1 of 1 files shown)


poiru:
Thank you for the patch, but this was done in Redis to secure their sandbox. I don't think this serves any purpose for Rainmeter because there is no sandbox. Lua scripts have access to `SKIN:Bang` and `os.execute` so they can execute arbitrary commands.

PR #401, opened 2025-04-03
Added the feature of GetOption

NSTechBytes:
# GetOption

Added the function of GetOption from skin plugin.

### DIFF
--- a/Build/DllExporter/DllExporter.csproj
+++ b/Build/DllExporter/DllExporter.csproj
@@ -10,7 +10,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>DllExporter</RootNamespace>
     <AssemblyName>DllExporter</AssemblyName>
-    <TargetFrameworkVersion>v4.7.1</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.8</TargetFrameworkVersion>
     <FileAlignment>512</FileAlignment>
     <PublishUrl>publish\</PublishUrl>
     <Install>true</Install>
@@ -29,6 +29,7 @@
     <BootstrapperEnabled>true</BootstrapperEnabled>
     <OutputPath Condition=" '$(Platform)' == 'x86' ">$(SolutionDir)x32-$(Configuration)\Tools\</OutputPath>
     <OutputPath Condition=" '$(Platform)' == 'x64' ">$(SolutionDir)x64-$(Configuration)\Tools\</OutputPath>
+    <TargetFrameworkProfile />
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|x86' ">
     <PlatformTarget>x86</PlatformTarget>
@@ -77,6 +78,9 @@
   <ItemGroup>
     <Compile Include="Program.cs" />
     <Compile Include="AssemblyInfo.cs" />
+  </ItemGroup>
+  <ItemGroup>
+    <None Include="app.config" />
   </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
 </Project>
--- a/Build/DllExporter/app.config
+++ b/Build/DllExporter/app.config
@@ -1,0 +1,3 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+<startup><supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.8"/></startup></configuration>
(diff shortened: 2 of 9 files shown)


Library/Export.cpp
```diff
@@ -16,6 +16,34 @@
 
 static std::wstring g_Buffer;
 
+LPCWSTR __stdcall RmGetOption(void* rm, LPCWSTR section, LPCWSTR option, LPCWSTR defValue, BOOL replaceMeasures)
+{
+	// Validate pointers
+	if (section == nullptr) { section = L""; }
```
poiru, line 22: Can you follow the convention of the existing code for null-checking, see `NULLCHECK` use in `RmReadString` for example?

Library/Export.cpp
```diff
@@ -16,6 +16,34 @@
...
 static std::wstring g_Buffer;
 
+LPCWSTR __stdcall RmGetOption(void* rm, LPCWSTR section, LPCWSTR option, LPCWSTR defValue, BOOL replaceMeasures)
+{
+	// Validate pointers
+	if (section == nullptr) { section = L""; }
+	if (option == nullptr) { option = L""; }
+	if (defValue == nullptr) { defValue = L""; }
+
+	// Cast the measure pointer
```
poiru, line 26: Please remove all comments in this function. Comments are meant to provide context or information that is not immediately visible from the code. These comments just rephrase what the code is doing in English, but in this case the code is very straightforward and not difficult to understand.

Please read this post for more information on this thinking: https://blog.codinghorror.com/code-tells-you-how-comments-tell-you-why/

Build/VS/Rainmeter.Cpp.props
```diff
@@ -27,7 +27,7 @@
   <!-- Macros for unit testing. -->
   <PropertyGroup Label="UserMacros">
     <!-- This is set to true by Build.bat when building full builds. -->
-    <ExcludeTests>false</ExcludeTests>
+    <ExcludeTests>true</ExcludeTests>
```
poiru, line 30: Why was this changed?

Plugins/API/RainmeterAPI.cs
```diff
@@ -39,6 +39,22 @@ static public implicit operator API(IntPtr rm)
...
+        /// <summary>
+        /// Retrieves the option defined in a specific section of the skin file and converts a relative path to an absolute path.
+        /// </summary>
+        /// <param name="section">Section name to be read from</param>
+        /// <param name="option">Option name to be read from the specified section</param>
+        /// <param name="defValue">Default value for the option if it is not found or invalid</param>
+        /// <param name="replaceMeasures">If true, replaces section variables in the returned string</param>
+        /// <returns>Returns the option value as a string</returns>
+        public static string GetOption(IntPtr rm, string section, string option, string defValue, bool replaceMeasures) {
+          IntPtr ptr = RmGetOption(rm, section, option, defValue, replaceMeasures ? 1 : 0);
```
poiru, line 54: What will happen if this is used with an older version of Rainmeter? Can you double-check what happens? Does the `LoadLibrary` call in Rainmeter fail for the DLL or does this cause a crash?

poiru (commented):
Hi @NSTechBytes, thanks for the PR. There are some possible issues so I reverted it. Could you please check the comments and open a new PR with them addressed?

Build/DllExporter/DllExporter.csproj
```diff
@@ -78,5 +79,8 @@
     <Compile Include="Program.cs" />
     <Compile Include="AssemblyInfo.cs" />
   </ItemGroup>
+  <ItemGroup>
+    <None Include="app.config" />
+  </ItemGroup>
   <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
 </Project>
```
brianferguson, line 86: Are any of these changes in this file even needed?

Build/DllExporter/app.config
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+<startup><supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.8"/></startup></configuration>
```
brianferguson, line 3: Like my comment above, is this file even needed?

Plugins/PluginUsageMonitor/PluginUsageMonitor.csproj
```diff
@@ -29,6 +29,7 @@
     <IsWebBootstrapper>false</IsWebBootstrapper>
     <UseApplicationTrust>false</UseApplicationTrust>
     <BootstrapperEnabled>true</BootstrapperEnabled>
+    <TargetFrameworkProfile />
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|x86' ">
     <DebugSymbols>true</DebugSymbols>
```
brianferguson, line 35: Same as comment above.

NSTechBytes: > Same as comment above.

Please see new PR https://github.com/rainmeter/rainmeter/pull/408

PR #408, opened 2025-04-04
Extended the SectionVariable and Added the GetOption for plugin api

NSTechBytes:
Extended the SectionVariable and Added the GetOption for plugin api.

### DIFF
--- a/Library/ConfigParser.cpp
+++ b/Library/ConfigParser.cpp
@@ -210,10 +210,10 @@ const std::wstring* ConfigParser::GetVariableOriginalName(const std::wstring& st
 }
 
 /*
-** Gets the value of a section variable. Returns true if strValue is set.
-** The selector is stripped from strVariable.
-**
-*/
+ ** Gets the value of a section variable. Returns true if strValue is set.
+ ** The selector is stripped from strVariable.
+ **
+ */
 bool ConfigParser::GetSectionVariable(std::wstring& strVariable, std::wstring& strValue, void* logEntry)
 {
 	if (!m_Skin) return false;
@@ -264,6 +264,13 @@ bool ConfigParser::GetSectionVariable(std::wstring& strVariable, std::wstring& s
 			}
 			else
 			{
+				
+				const std::wstring optionValue = m_Skin->GetParser().ReadString(strVariable.c_str(), selector.c_str(), L"");
+				if (!optionValue.empty())
+				{
+					strValue = optionValue;
+					return true;
+				}
 				return false;
 			}
 
@@ -279,7 +286,7 @@ bool ConfigParser::GetSectionVariable(std::wstring& strVariable, std::wstring& s
 	// EscapeRegExp: [Measure:EscapeRegExp] (Escapes regular expression syntax, used for 'IfMatch')
 	// EncodeUrl: [Measure:EncodeUrl] (Escapes URL reserved characters)
 	// TimeStamp: [TimeMeasure:TimeStamp] (ONLY for Time measures, returns the Windows timestamp of the measure)
-
+	
 	// Script: [ScriptMeasure:SomeFunction()], [ScriptMeasure:Something('Something')]
 	// NOTE: Parenthesis are required. Arguments enclosed in single or double quotes are treated as strings, otherwise
 	//   they are treated as numbers. If the lua function returns a number, it will be converted to a string.
@@ -333,15 +340,29 @@ bool ConfigParser::GetSectionVariable(std::wstring& strVariable, std::wstring& s
 			const auto type = measure->GetTypeID();
 			if (type == TypeID<MeasureScript>())
 			{
-				valueType = ValueType::Script;  // Needed?
+				valueType = ValueType::Script;  // Needed?  
 				MeasureScript* script = (MeasureScript*)measure;
 				retValue = script->CommandWithReturn(selectorSz, strValue, logEntry);
 			}
 			else if (type == TypeID<MeasurePlugin>())
 			{
-				valueType = ValueType::Plugin;  // Needed?
+				valueType = ValueType::Plugin;  // Needed?  
 				MeasurePlugin* plugin = (MeasurePlugin*)measure;
 				retValue = plugin->CommandWithReturn(selectorSz, strValue, logEntry);
+			}
+			else
+			{
+				
+				const std::wstring optionValue = m_Skin->GetParser().ReadString(strVariable.c_str(), selector.c_str(), L"");
+				if (!optionValue.empty())
+				{
+					strValue = optionValue;
+					retValue = true;
+				}
+				else
+				{
+					retValue = false;
+				}
 			}
 
 			m_StyleTemplate = meterStyle;
@@ -374,7 +395,7 @@ bool ConfigParser::GetSectionVariable(std::wstring& strVariable, std::wstring& s
 				}
 
 				strVariable.resize(colonPos);
-			}
+			} 
 			while (0);
 		}
 	}
@@ -445,7 +466,7 @@ bool ConfigParser::GetSectionVariable(std::wstring& strVariable, std::wstring& s
 			(valueType == ValueType::Percentual) ? measure->GetRelativeValue() * 100.0 :
 			(valueType == ValueType::Max)        ? measure->GetMaxValue() / scale :
 			(valueType == ValueType::Min)        ? measure->GetMinValue() / scale :
-			                                       measure->GetValue() / scale;
+			measure->GetValue() / scale;
 		int decimals = 10;
 		if (decimalsSz)
 		{
@@ -467,7 +488,7 @@ bool ConfigParser::GetSectionVariable(std::wstring& strVariable, std::wstring& s
 		WCHAR buffer[128] = { 0 };
 		_snwprintf_s(format, _TRUNCATE, L"%%.%if", decimals);
 		int bufferLen = _snwprintf_s(buffer, _TRUNCATE, format, value);
-			
+
 		if (!decimalsSz)
 		{
 			// Remove trailing zeros if decimal count was not specified.
@@ -478,7 +499,7 @@ bool ConfigParser::GetSectionVariable(std::wstring& strVariable, std::wstring& s
 		strValue.assign(buffer, bufferLen);
 		return true;
 	}
-	
+
 	return false;
 }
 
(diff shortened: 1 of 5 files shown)


Library/ConfigParser.cpp
```diff
@@ -210,10 +210,10 @@ const std::wstring* ConfigParser::GetVariableOriginalName(const std::wstring& st
...
 
 /*
-** Gets the value of a section variable. Returns true if strValue is set.
-** The selector is stripped from strVariable.
-**
-*/
+ ** Gets the value of a section variable. Returns true if strValue is set.
+ ** The selector is stripped from strVariable.
+ **
+ */
```
poiru, line 216: Please revert these spacing changes. The Rainmeter commenting style is weird (I don't like it either), but lets keep it consistent.

Library/ConfigParser.cpp
```diff
@@ -279,7 +286,7 @@ bool ConfigParser::GetSectionVariable(std::wstring& strVariable, std::wstring& s
 	// EscapeRegExp: [Measure:EscapeRegExp] (Escapes regular expression syntax, used for 'IfMatch')
 	// EncodeUrl: [Measure:EncodeUrl] (Escapes URL reserved characters)
 	// TimeStamp: [TimeMeasure:TimeStamp] (ONLY for Time measures, returns the Windows timestamp of the measure)
-
+
```
poiru, line 289: Please revert spacing changes like this here and elsewhere in this file.

Library/ConfigParser.cpp
```diff
@@ -264,6 +264,13 @@ bool ConfigParser::GetSectionVariable(std::wstring& strVariable, std::wstring& s
 			}
 			else
 			{
+				
+				const std::wstring optionValue = m_Skin->GetParser().ReadString(strVariable.c_str(), selector.c_str(), L"");
```
poiru, line 268: I'm not sure what you would gain from making this change. Why or where would you want to read an option from a meter/measure? (This is where a comment describing your change would be helpful in the PR description!)

poiru:
Hey @NSTechBytes,

I'm sorry if my request seems unreasonable. I can understand your frustration because I had similar moments when I started learning C++ 16 years in order to contribute to Rainmeter. But in the end, I learned a ton from the more experienced developers and it served me very well in my career.

I would be happy to help you how to address the spacing issue so that we can focus on the actual change you want to make. Git actually makes this very easy to avoid (and also fix it now) and I'm happy to take the time to explain how to use Git to solve this. Let me know if that sounds interesting to you — otherwise we will unfortunately have to close this PR.

jsmorley:
>Thanks for the story, but I’m not really here for a Git lesson. The code works fine as is—spaces aren’t hurting anything. If that’s a dealbreaker, close the PR. I’m not sweating it.

Well, we can just close it and forget it then. We have standards for our code, which anyone who has done any coding professionally will understand has enormous benefits in the long run. 

We don't expect someone new to our code to do it exactly like we do it from the start, but if you can't accept any advice on how to make your code fit our structure, then we can't make progress.

This is a damn shame. Very unexpected, giving how well you reacted to suggestions and such on the forums.

Yincognyto:
Not that my opinion would matter, but I think we can't see the forest for the trees here. Having this PR in the Rainmeter code itself and not just in a plugin has major benefits for future plugin developers. Seriously now, the issue of both having spaces or not is a really unfortunate - to put it mildly - way to hinder an useful addition to Rainmeter.

While RicardoTM has some C# experience that I lack, we're both pretty ignorant when it comes to Git, but can't one of us just upload the file somewhere with the spaces issue solved, so we can implement this? Come on now, all of you, don't be overly pedantic or emotional to let little things sabotage a bigger goal...

poiru:
@Yincognyto We're not being pedantic or emotional. The spacing issue is trivial as you said and can easily be resolved. But these questions have yet to be addressed:

* https://github.com/rainmeter/rainmeter/pull/401#discussion_r2028493147
* https://github.com/rainmeter/rainmeter/pull/408#discussion_r2028834492

poiru:
@Yincognyto The spacing issue is trivial as you said and can easily be resolved. But these questions have yet to be addressed:

* https://github.com/rainmeter/rainmeter/pull/401#discussion_r2028493147
* https://github.com/rainmeter/rainmeter/pull/408#discussion_r2028834492

>Having this PR in the Rainmeter code itself and not just in a plugin has major benefits for future plugin developers.

It is not possible to have this change only in a plugin. Rainmeter needs to provide the API for it to use, but we cannot do that in a way that breaks things.

NSTechBytes:
Hey, I’m sorry for how I came off earlier. I got defensive and took it too personally—didn’t mean to snap like that. I get it, consistency matters, especially with a team project like this. I’m still figuring out how you guys do things, and I shouldn’t have pushed back so hard. Can we keep the PR open? I’ll tweak the spacing to match your standards. Appreciate your patience with me on this.

poiru:
>it looks more like a question on very purpose of the PR rather than something that could break things

@Yincognyto Right, but as a non-active Rainmeter developer, we need context on why this is useful and why this is done. I can see what the section variable code changes are doing, but I don't know why you would want this.

For example, the code as it is implemented now creates inconsistencies. Lets take this example:
```ini
[Meter]
W=(100*2)
ClipStringW=(100*2)
```

The previously supported  `[Meter:W]` would return the actual width as the number so `200`. But `[Meter:ClipStringW]` would return `(100*2)`. This is not good design.

If the intention would have been explained, it would have been easier for me to propose an alternative way to achieve what you want. I'm not an active developer (we currently don't have any active developers), so it is very much appreciated that new code change proposals come with enough context to explain WHY it is needed rather than WHAT it it is changing. The WHAT is obvious most of the time from the code itself, but the WHY requires us to understand your thinking.

In this case, you probably would want to introduce a totally new way to access the original string regardless of whether it was a formula or whatever. For example, `[Meter:Raw:X]` and `[Meter:Raw:ClipStringW]` could both return `(100*2)`.

I don't know if "Raw" is the best term to use here (or whether it should be a symbol), but perhaps @jsmorley can help with deciding that.

>Hey, I’m sorry for how I came off earlier. I got defensive and took it too personally—didn’t mean to snap like that. I get it, consistency matters, especially with a team project like this. I’m still figuring out how you guys do things, and I shouldn’t have pushed back so hard. Can we keep the PR open? I’ll tweak the spacing to match your standards. Appreciate your patience with me on this.

@NSTechBytes No worries, I understand. Please see the comment above and let me know what you think.

Library/ConfigParser.cpp
NSTechBytes: Hey, about your note to 'revert spacing changes like this here and elsewhere in this file'—do you mean I should just undo the spacing edits I added, or is there a specific way you want it formatted to match the standard? Just want to get it right.

NSTechBytes:
Hey, about your note to 'revert spacing changes like this here and elsewhere in this file'—do you mean I should just undo the spacing edits I added, or is there a specific way you want it formatted to match the standard? Just want to get it right.

poiru:
> Hey, about your note to 'revert spacing changes like this here and elsewhere in this file'—do you mean I should just undo the spacing edits I added, or is there a specific way you want it formatted to match the standard? Just want to get it right.

Yep, please undo them. Not just because of the inconsistency, but because it obscures the actual point of this PR.

A good PR is focused one one thing to make it easy to review, but this PR is doing 3 things: section variable change, GetOption API, and the spacing. So it would be best to undo the spacing change and then have two separate PRs for the GetOption API and section variable change. The GetOption API should be fine as-is to merge if we can confirm what happens when a new plugin is loaded in old Rainmeter for both a C# and C++ plugin. But the section variable stuff needs more work so it would be best to have it in a separate PR.

You would need to create separate Git branches for that — right now it seems like you are using the `master` branch for everything.

Yincognyto:
> > it looks more like a question on very purpose of the PR rather than something that could break things
> 
> @Yincognyto Right, but as non-active Rainmeter developers, we need context on why this is useful and why this is done the way it is. I can see what the section variable code changes are doing, but I don't know why you would want this.
> 
> For example, the code as it is implemented now creates inconsistencies. Lets take this example:
> 
> ```ini
> [Meter]
> W=(100*2)
> ClipStringW=(100*2)
> ```
> 
> The previously supported `[Meter:W]` would return the actual width as the number so `200`. But `[Meter:ClipStringW]` would return `(100*2)`. This is not good design.
> 
> If the intention would have been explained, it would have been easier for me to propose an alternative way to achieve what you want. In this case, you probably would want to introduce a totally new way to access the original string regardless of whether it was a formula or whatever. For example, `[Meter:Raw:X]` and `[Meter:Raw:ClipStringW]` could both return `(100*2)`.
> 
> I don't know if "Raw" is the best term to use here (or whether it should be a symbol), but perhaps @jsmorley can help with deciding that.

Yep, I understand, it's about the 'literal' vs 'resolved' value (which could benefit a way to choose between them, if it ever gets to that). As for why would anyone want any of them, I think it's pretty obvious: some skin designers would need the 'resolved' value in order to use it further, other skin designers would need the 'literal' value for further processing e.g. substitutions and such. But yeah, returning the same type of value whether it's [Meter:W] or [Meter:AnotherOption] would be the correct and consistent way to do this. Again, ideally, it would be nice if you could choose which type you want (like you can do in Lua with the bReplaceMeasures argument, for example).

Other than that, great to see things are back in track with the PR - well done everyone. A little patience helps to get a result that suits all raised considerations. ;)

jsmorley:
It has been suggested that we support something like:

[Meter:X:AsNumber] or [Meter:X:AsString]

jsmorley:
One other question I would raise is; If we are going to change Rainmeter to support this "call" from a plugin, should we consider adding the functionality to the Lua "glue" as well?

Yincognyto:
> One other question I would raise is; If we are going to change Rainmeter to support this "call" from a plugin, should we consider adding the functionality to the Lua "glue" as well?

Personally, I'd definitely do it (inline usage is convenient when it comes to Lua). That being said, Lua already has some functions in that regard, so I'm not sure how to integrate things. I think Lua should generally have at least what the .ini code is capable of.

Even though these things not necessarily related, ideally, from my point of view, would have been, say, a way to be able to use a `table['Skin'|N]['Section'|N]['Option'|N]`, that one can look up into in any way he needs for skins, sections, options and such.

This is strictly from a skin designer point of view. From a program or plugin developer point of view, things might be different.

PR #409, opened 2025-04-05
Added the Option of GetOption for plugin Usage.

NSTechBytes:
Hey, I just want to say sorry again for my earlier reaction. I really appreciate your understanding and support—it helps more than you know. About the PR title, I went with "Added the Option of GetOption for Plugin Usage". Do you think that works, or would you like me to tweak it? I’d love to hear your thoughts to make sure it’s spot on.

### DIFF
--- a/Library/Export.cpp
+++ b/Library/Export.cpp
@@ -247,3 +247,27 @@ LPCWSTR PluginBridge(LPCWSTR command, LPCWSTR data)
 
 	return L"noop";
 }
+
+LPCWSTR __stdcall RmGetOption(void* rm, LPCWSTR section, LPCWSTR option, LPCWSTR defValue, BOOL replaceMeasures)
+{
+
+	if (section == nullptr) { section = L""; }
+	if (option == nullptr) { option = L""; }
+	if (defValue == nullptr) { defValue = L""; }
+
+	MeasurePlugin* measure = (MeasurePlugin*)rm;
+	ConfigParser& parser = measure->GetSkin()->GetParser();
+
+	const std::wstring& style = parser.ReadString(section, L"MeterStyle", L"");
+	if (!style.empty())
+	{
+		parser.SetStyleTemplate(style);
+	}
+
+	const std::wstring& value = parser.ReadString(section, option, defValue, replaceMeasures != FALSE);
+	g_Buffer = value;
+
+	parser.ClearStyleTemplate();
+
+	return g_Buffer.c_str();
+}
--- a/Plugins/API/RainmeterAPI.cs
+++ b/Plugins/API/RainmeterAPI.cs
@@ -16,6 +16,17 @@ namespace Rainmeter
     public class API
     {
         private IntPtr m_Rm;
+        private static bool _rmGetOptionExists;
+
+        static API()
+        {
+            IntPtr hModule = LoadLibrary("Rainmeter.dll");
+            if (hModule != IntPtr.Zero)
+            {
+                IntPtr procAddress = GetProcAddress(hModule, "RmGetOption");
+                _rmGetOptionExists = (procAddress != IntPtr.Zero);
+            }
+        }
 
         public API(IntPtr rm)
         {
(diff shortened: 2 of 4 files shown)


jsmorley:
We would need a new PR for the Plugin SDK as well?

Library/Export.cpp
```diff
@@ -247,3 +247,27 @@ LPCWSTR PluginBridge(LPCWSTR command, LPCWSTR data)
 
 	return L"noop";
 }
+
+LPCWSTR __stdcall RmGetOption(void* rm, LPCWSTR section, LPCWSTR option, LPCWSTR defValue, BOOL replaceMeasures)
+{
+
+	if (section == nullptr) { section = L""; }
+	if (option == nullptr) { option = L""; }
+	if (defValue == nullptr) { defValue = L""; }
```
poiru, line 256: Can we use `NULLCHECK` here?

Plugins/API/RainmeterAPI.h
```diff
@@ -179,6 +179,20 @@ LIBRARY_EXPORT void __cdecl RmLogF(void* rm, int level, LPCWSTR format, ...);
...
+/// <param name="rm">Pointer to the plugin measure</param>
+/// <param name="section">Section name to be read from</param>
+/// <param name="option">Option name to be read from the specified section</param>
+/// <param name="defValue">Default value if the option is not found or invalid</param>
+/// <param name="replaceMeasures">If true, replaces section variables in the returned string</param>
+/// <returns>Returns the option value as a string (LPCWSTR)</returns>
+#ifdef __cplusplus
+LIBRARY_EXPORT LPCWSTR __stdcall RmGetOption(void* rm, LPCWSTR section, LPCWSTR option, LPCWSTR defValue, BOOL replaceMeasures = TRUE);
+#else
+LIBRARY_EXPORT LPCWSTR __stdcall RmGetOption(void* rm, LPCWSTR section, LPCWSTR option, LPCWSTR defValue, BOOL replaceMeasures);
```
poiru, line 194: Please remove this entirely and lets instead rename `RmGetOptionSafe` to be called `RmGetOption`. That way it will always be safe, there is no need to expose the unsafe version here.

Plugins/API/RainmeterAPI.h
```diff
@@ -205,6 +219,39 @@ __inline LPCWSTR RmReadPath(void* rm, LPCWSTR option, LPCWSTR defValue)
 	return RmPathToAbsolute(rm, relativePath);
 }
 
+/// <summary>
+/// Safely retrieves the option defined in a specific section of the skin file.
+/// If RmGetOption is not available (e.g., in older Rainmeter versions), returns the default value.
```
poiru, line 224: I would rephrase this like so:
```suggestion
/// In older Rainmeter versions where RmGetOption is not available, always returns the default value.
```

Plugins/API/RainmeterAPI.h
```diff
@@ -205,6 +219,39 @@ __inline LPCWSTR RmReadPath(void* rm, LPCWSTR option, LPCWSTR defValue)
...
+        if (rainmeterDll != nullptr) {
+            rmGetOption = (RmGetOptionFunc)GetProcAddress(rainmeterDll, "RmGetOption");
+        }
+    }
+    if (rmGetOption != nullptr) {
+        return rmGetOption(rm, section, option, defValue, replaceMeasures);
+    } else {
+        return defValue;
+    }
+}
```
poiru, line 254: I fixed the code style (tabs instead of spaces, `{` on its own line). The code can also be simplified a lot by avoiding unnecessary error-checking - we only need to check it in one place since the others can fail without issue.

```suggestion
inline LPCWSTR RmGetOption(void* rm, LPCWSTR section, LPCWSTR option, LPCWSTR defValue, BOOL replaceMeasures = TRUE)
{
	typedef LPCWSTR (__stdcall *RmGetOptionFunc)(void*, LPCWSTR, LPCWSTR, LPCWSTR, BOOL);
	static RmGetOptionFunc rmGetOption = (RmGetOptionFunc)GetProcAddress(GetModuleHandle(L"Rainmeter.dll"), "RmGetOption");

	if (rmGetOption)
	{
		return rmGetOption(rm, section, option, defValue, replaceMeasures);
	}

	return defValue;
}

```

Plugins/API/RainmeterAPI.cs
```diff
@@ -431,6 +442,49 @@ public void LogF(LogType type, string format, params Object[] args)
...
         }
+
+        [DllImport("kernel32.dll", SetLastError = true, CharSet = CharSet.Unicode)]
+        private static extern IntPtr LoadLibrary(string lpFileName);
+
+        [DllImport("kernel32.dll", SetLastError = true)]
+        private static extern IntPtr GetProcAddress(IntPtr hModule, string lpProcName);
+
+        [DllImport("Rainmeter.dll", CharSet = CharSet.Unicode, CallingConvention = CallingConvention.StdCall, EntryPoint = "RmGetOption")]
+        private static extern IntPtr GetOptionStatic(IntPtr rm, string section, string option, string defValue, int replaceMeasures);
```
poiru, line 453: Does this actually work? Did you test a new C# plugin using `GetOption` on an older Rainmeter version?

poiru (commented):
Thank you, looks a lot better. Left a few comments!

For the title, I think something like "Add RmGetOption to plugin API" would be more descriptive and clear.

NSTechBytes:
Sure, I will add a new PR after Poiru’s review.

RicardoTM05:
Please consider my review as well, it is important that the function follows the same syntax as the other functions.

`public string GetOption(string section, string option, string defValue, bool replaceMeasures = true)`

Plugins/API/RainmeterAPI.cs
```diff
@@ -431,6 +442,49 @@ public void LogF(LogType type, string format, params Object[] args)
...
+        ///{
+        ///    Measure measure = data;
+        ///    Rainmeter.API api = (Rainmeter.API)rm;
+        ///    string sectionName = api.ReadString("SectionName", "MySection");
+        ///    string keyOption = api.ReadString("KeyOption", "MyOption");
+        ///    string value = API.GetOption(rm, sectionName, keyOption, "DefaultValue", true);
+        ///}
+        /// </code>
+        /// </example>
+        public static string GetOption(IntPtr rm, string section, string option, string defValue, bool replaceMeasures)
```
poiru, line 478: Please use `m_Rm` instead of taking `rm` as a parameter like the other functions.

Plugins/API/RainmeterAPI.cs
```diff
@@ -431,6 +442,49 @@ public void LogF(LogType type, string format, params Object[] args)
...
             RmLog(this.m_Rm, type, string.Format(format, args));
         }
+
+        [DllImport("kernel32.dll", SetLastError = true, CharSet = CharSet.Unicode)]
+        private static extern IntPtr LoadLibrary(string lpFileName);
+
+        [DllImport("kernel32.dll", SetLastError = true)]
+        private static extern IntPtr GetProcAddress(IntPtr hModule, string lpProcName);
+
+        [DllImport("Rainmeter.dll", CharSet = CharSet.Unicode, CallingConvention = CallingConvention.StdCall, EntryPoint = "RmGetOption")]
```
poiru, line 452: @NSTechBytes Please confirm that you tested this. You marked this as resolved, but I want to you to explicitly say that it was tested with an old version of Rainmeter :)

I don't think this works like the C++ version because you are importing `RmGetOption` instead of accessing it through `GetProcAddress`.

Plugins/API/RainmeterAPI.cs
```diff
@@ -431,6 +442,49 @@ public void LogF(LogType type, string format, params Object[] args)
...
+        ///    Rainmeter.API api = (Rainmeter.API)rm;
+        ///    string sectionName = api.ReadString("SectionName", "MySection");
+        ///    string keyOption = api.ReadString("KeyOption", "MyOption");
+        ///    string value = API.GetOption(rm, sectionName, keyOption, "DefaultValue", true);
+        ///}
+        /// </code>
+        /// </example>
+        public static string GetOption(IntPtr rm, string section, string option, string defValue, bool replaceMeasures)
+        {
+            if (!_rmGetOptionExists)
```
poiru, line 480: Instead of the `_rmGetOptionExists` approach, you can simply try to call the real `RmReadOption` function and then handle the `EntryPointNotFoundException` for the cases where it is not available.

NSTechBytes:
I’ll take a look at it soon.

PR #413, opened 2025-04-09
Fix 'Show Desktop' hiding skins on Windows 11 24H2

### LINKED ISSUE #377: On Windows 11 24H2: showing desktop will hide rainmeter widgets
On Windows 11 24H2 (Build 26xxx), showing desktop by pressing Win+D or clicking the right corner of taskbar will cause all rainmeter widgets to disappear as well.

It affects all current 24h2 builds (Canary, Dev or Release Preview). 23H2 is not affected.

poiru:
Win+D and the 'Show Desktop' button were both hiding skins on Windows 11 24H2. We now account for the changes in the desktop WorkerW hierarchy and limit the change to recent builds of Windows 11 to limit affecting older operating systems.

Fixes #377

### DIFF
--- a/Library/System.cpp
+++ b/Library/System.cpp
@@ -662,6 +662,32 @@ HWND System::GetDefaultShellWindow()
 	return ShellW;
 }
 
+// Windows 11 24H2 reordered the the desktop windows.
+//
+// Spy++ output before Windows 11 24H2:
+//
+//   0x00010190 "" WorkerW
+//     ...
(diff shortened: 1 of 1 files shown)


Library/System.cpp
```diff
@@ -662,6 +662,32 @@ HWND System::GetDefaultShellWindow()
 	return ShellW;
 }
 
+// Windows 11 24H2 reordered the the desktop windows.
```
raziel-, line 665: "the the"

PR #428, opened 2026-02-07
Visibility-based Auto-pause GameMode (draft)

roj234:
What is Visibility-based Auto-pause GameMode:
- When you can't see desktop, pause all Meters
- Vice versa
- Read function `IsDesktopCoveredOnMonitor` for details.

This is a draft because
- I'm not familiar with C++
- I'm **VERY** not familiar with how to save/load config in C++
- I'm **VERY** not familiar with how to display GUI in C++

As a result:
- This PR did not add any buttons or ini options  
- This PR is a hack: it changed old behaviour  
- Corner case is not covered: current code only check one monitor that POINT(0, 0) owns to

### DIFF
--- a/Library/GameMode.cpp
+++ b/Library/GameMode.cpp
@@ -15,6 +15,7 @@
 #include "Rainmeter.h"
 #include "System.h"
 #include "../Common/StringUtil.h"
+#include "IsDesktopVisible.hpp"
 
 namespace
 {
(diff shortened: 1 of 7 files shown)


poiru:
We have implemented this functionality in a0fc0b0 although it is not hooked up to GameMode yet.

PR #442, opened 2026-05-23
Re-read skin positions on display change

### LINKED ISSUE #441: Skin positions stay clamped after display topology changes
When Rainmeter starts while one monitor is unavailable, skins can be clamped to the available monitor. After the missing monitor is turned on again, the skins stay at the clamped in-memory positions until Rainmeter is refreshed manually.

Repro:
1. Place skins on a secondary monitor.
2. Start Windows/Rainmeter with another monitor turned off, so the active monitor topology differs from the saved layout.
3. Turn the missing monitor back on.

Expected:
Skins return to their saved positions after Rainmeter handles the display change.

Actual:
Some skins remain at the temporary clamped positions until manual refresh.

xob0t:
Fixes #441.

When a display change triggers a delayed skin move, re-read the saved WindowX/WindowY/AnchorX/AnchorY options before resizing/moving the skin. This makes the display-change path use the saved positions instead of any values temporarily clamped while a monitor was unavailable.

Tested:
- Manually verified with Rainmeter started while one monitor was unavailable, then restored the monitor and confirmed skins returned without manual refresh.


poiru:
Thanks for the PR and apologies for the delay! We've implemented this in e5b132a0 using a different approach, but the net result should be the same. Closing this PR.

PR #476, opened 2026-08-15
NowPlaying/Cover: Add .webp and .avif local cover support

DrSakuu:
Rainmeter supports displaying .webp and .avif images, but those extensions are missing when displaying a cover from a local file.

Should the order of extensions be rearranged to prioritize newer formats? Maybe `L"avif", L"webp", L"png", L"jpg", L"jpeg", L"bmp"` . Should it be changed to `LPCTSTR extName[]` and `i < _countof(extName)` to eliminate the `extCount` constant? Or even `const auto& i : extName`?


poiru:
Thanks! Will also implement your suggestions in a follow-up.

PR #480, opened 2026-08-20
Defer Rainmeter.ini writes from window-state bangs

### LINKED ISSUE #471: [!ZPos] The ZPos bang always saves its setting to Rainmeter.ini
# rainmeter/Library/Skin.cpp
```cpp
1350		case Bang::ZPos:
1351			SetWindowZPosition((ZPOSITION)m_Parser.ParseInt(args[0].c_str(), 0));
1352			break;

...

4343	void Skin::SetWindowZPosition(ZPOSITION zPos)
4344	{
4345		ChangeSingleZPos(zPos);
4346		WriteOptions(OPTION_ALWAYSONTOP);
4347	}
```

When the user calls !ZPos, it <ins>always</ins> results in a file write operation.  If this is the intended behaviour, it should be documented as such, otherwise, the user could be wearing down their SSD drive for no practical benefit.

On a related note, the _Manage Rainmeter_ UI is a bit misleading.  The Skins tab allows the user to configure:
- _Coordinates_: the left/top position of the skin
- _Position_: the Windows ZOrder position of the skin
- _Save position_: auto save "position"

Based on the UI terminology, the _Save position_ checkbox is more accurately _Save coordinates_.  Currently, there is no way to disable auto save of the ZOrder position.

Would argue that all skin-level, bang-related position operations should not auto save, and the user would call a separate bang to save if/when needed.  All other position operations (changes through _Manage Rainmeter_ UI, Rainmeter-level drag/drop/resize, etc) should be auto saved.

Mitanshp5:
Fixes #471

### Summary of Changes
Bangs that modify window/skin states (`!ZPos`, `!Draggable`, `!ClickThrough`, `!SnapEdges`, `!KeepOnScreen`, and `!AutoSelectScreen`) were calling setter methods that unconditionally wrote settings to `Rainmeter.ini` via `WriteOptions(...)`.

When skins use these bangs dynamically in fast update loops or interactive animations, this resulted in continuous, rapid disk writes and unnecessary SSD wear.

### Solution
1. **Scoped Deferral in `Skin::DoBang()`**: Set `m_DeferOptionWrites = true` at the start of `Skin::DoBang()` using `Scoped` to restore it on exit. This avoids altering any setter function signatures (`SetWindowZPosition`, `SetClickThrough`, etc.).
2. **Debounced Disk Writes in `WriteOptions()`**: In `Skin::WriteOptions()`, when `m_DeferOptionWrites` is active, option flags accumulate into `m_PendingWriteOptions` and schedule a 10-second debounce timer (`TIMER_WRITE_OPTIONS`), guarded by `m_WriteOptionsScheduled` so subsequent bangs do not repeatedly postpone the timer.
3. **Extracted `WriteDeferredOptions()` Flush Helper**: Centralized the timer cleanup and pending write flush into `WriteDeferredOptions()`, which is called upon:
   - Debounce timer expiration
   - Skin unload / Rainmeter shutdown (`Skin::Dispose`)
   - Skin refresh (`Skin::Refresh`)
   - Skin deselection (`Skin::Deselect`)
4. **Preserved Immediate Writes for UI & Menus**: Since `m_DeferOptionWrites` is only set during bang execution, all user changes made through the Manage Dialog and context menus continue to write immediately to disk as expected.

### DIFF
--- a/Library/Skin.cpp
+++ b/Library/Skin.cpp
@@ -29,6 +29,7 @@
 #include "../Version.h"
 #include "../Common/DpiUtil.h"
 #include "../Common/PathUtil.h"
+#include "../Common/ScopedFunction.h"
 #include "../Common/StringUtil.h"
 #include "../Common/Gfx/Util/EffectStream.h"
 
@@ -43,17 +44,19 @@ enum TIMER
 	TIMER_DEACTIVATE = 5,
 	TIMER_PREVENT_MOVE = 6,
 	TIMER_CARET = 7,
+	TIMER_WRITE_OPTIONS = 8,
 
 	// Update this when adding a new timer.
-	TIMER_MAX = 7
+	TIMER_MAX = 8
 };
 
 enum INTERVAL
 {
-	INTERVAL_METER      = 1000,
-	INTERVAL_MOUSE      = 500,
-	INTERVAL_FADE       = 10,
-	INTERVAL_TRANSITION = 100
+	INTERVAL_METER         = 1000,
+	INTERVAL_MOUSE         = 500,
+	INTERVAL_FADE          = 10,
+	INTERVAL_TRANSITION    = 100,
+	INTERVAL_WRITE_OPTIONS = 10000
 };
 
 int Skin::c_InstanceCount = 0;
@@ -73,6 +76,8 @@ Skin::Skin(const std::wstring& folderPath, const std::wstring& file, const bool 
 	m_SelectionOverlay(),
 	m_DropTarget(),
 	m_PendingWriteOptions(0),
+	m_DeferOptionWrites(false),
+	m_WriteOptionsScheduled(false),
 	m_SuspendResumeNotification(nullptr),
 	m_Mouse(this),
 	m_MouseOver(false),
@@ -203,6 +208,8 @@ void Skin::Dispose(bool refresh)
 	KillTimer(m_Window, TIMER_TRANSITION);
 	KillTimer(m_Window, TIMER_PREVENT_MOVE);
 	KillTimer(m_Window, TIMER_CARET);
+
+	WriteDeferredOptions();
 
 	m_ActiveFade = false;
 	m_FadeStartTime = 0;
@@ -462,11 +469,7 @@ void Skin::Refresh(bool init, bool all)
 
 	LogNoticeF(this, L"Refreshing skin");
 
-	if (m_PendingWriteOptions != 0)
-	{
-		WriteOptions(m_PendingWriteOptions);
-		m_PendingWriteOptions = 0;
-	}
+	WriteDeferredOptions();
 
 	SetResizeWindowMode(RESIZEMODE_RESET);
 
@@ -943,11 +946,7 @@ void Skin::Deselect()
 
 	m_SelectionOverlay.reset();
 
-	if (m_PendingWriteOptions != 0)
-	{
-		WriteOptions(m_PendingWriteOptions);
-		m_PendingWriteOptions = 0;
-	}
+	WriteDeferredOptions();
 
 	for (const auto& meter : m_Meters) meter->ResetToolTip();
 
@@ -1074,6 +1073,10 @@ void Skin::ChangeSingleZPos(ZPOSITION zPos, bool all)
 // Correct number of arguments must be passed (or use Rainmeter::ExecuteBang).
 void Skin::DoBang(Bang bang, const std::vector<std::wstring>& args)
 {
+	const bool oldDeferOptionWrites = m_DeferOptionWrites;
+	m_DeferOptionWrites = true;
+	auto deferScope = Scoped([&] { m_DeferOptionWrites = oldDeferOptionWrites; });
+
 	switch (bang)
 	{
 	case Bang::Refresh:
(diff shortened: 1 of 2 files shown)


poiru:
Hey! Thanks for the PR 🙏

I think this approach is not ideal because changes will be lost when quitting Rainmeter. I think we should defer these writes using the existing `m_PendingWriteOptions` approach so that changes do eventually get written out. In this case, we could defer the writes for 10 seconds, for example.

I'm happy to implement that later, but if you'd like to give it a shot, please let me know and I'll hold off!

Mitanshp5:
Hey @poiru , thanks for the feedback — that makes sense, I hadn't accounted for the case where Rainmeter quits before the next write happens, so settings would just be silently lost. Good catch.
I'd like to take a shot at implementing the deferred write with m_PendingWriteOptions, using a ~10s debounce as you suggested. Just to make sure I get the approach right before I start:
Should the deferred write also flush immediately on skin unload / Rainmeter shutdown, so pending state isn't lost on quit?
Is there an existing timer/scheduler pattern in the codebase you'd want me to reuse for the debounce, or should I add a new one scoped to Skin?
I'll hold off touching anything else until I hear back — let me know if that sounds right and I'll get started.

poiru:
>Should the deferred write also flush immediately on skin unload / Rainmeter shutdown, so pending state isn't lost on quit?

Yep!

>Is there an existing timer/scheduler pattern in the codebase you'd want me to reuse for the debounce, or should I add a new one scoped to Skin?

You could do something like this:

https://github.com/rainmeter/rainmeter/blob/7d05a7b661f92fe31cf9ddb3d178a185d1b653bb/Library/Skin.cpp#L4720-L4727

That will replace the timer if it already exists, which is not exactly what we want in this case since it would keep deferring the write if there are more changes before the timeout is up. You'll want to avoid replacing the timer by keeping a flag (e.g. `m_WriteOptionsScheduled`).

Mitanshp5:
Hey @poiru, I've updated the PR with the deferred write implementation:
- Debounced bang-triggered writes using `m_PendingWriteOptions` and a 10s timer (`TIMER_WRITE_OPTIONS`), guarded by `m_WriteOptionsScheduled` to prevent resetting the timer on rapid updates.
- Handled flushing and timer cleanup on skin unload/quit (`Skin::Dispose`), refresh (`Skin::Refresh`), and deselect (`Skin::Deselect`).
- Added a `force` parameter to `WriteOptions` so lifecycle flushes reliably write to disk even if the skin is currently selected.
- Kept immediate writes for UI and context menu actions.

Let me know if you'd like any further adjustments!

Library/Skin.cpp
```diff
@@ -2277,14 +2291,14 @@ void Skin::ReadOptions(ConfigParser& parser, LPCWSTR section, bool isDefault)
...
+void Skin::WriteOptions(INT setting, bool force)
 {
 	if (setting & OPTION_POSITION)
 	{
 		ComputeOptionValueFromPosition();
 	}
 
-	if (IsSelected())
+	if (!force && IsSelected())
 	{
```
poiru, line 2291: Instead of `force` parameter, you could add a `m_DeferOptionWrites` member and set it to `true` at the top of `Skin::DoBang()` with something like this:

```cpp
const bool oldDeferOptionWrites = m_DeferOptionWrites;
m_DeferOptionWrites = true;
auto deferScope = Scoped([&] { m_DeferOptionWrites = oldDeferOptionWrites; });
```

Then you could just check for `m_DeferOptionWrites` here instead of `force`. It would allow you to revert the changes to `Skin::SetClickThrough` etc., remove `DeferWriteOptions()`, and move the timer creation logic into this block.

Library/Skin.cpp
```diff
@@ -203,6 +205,14 @@ void Skin::Dispose(bool refresh)
...
 	KillTimer(m_Window, TIMER_PREVENT_MOVE);
 	KillTimer(m_Window, TIMER_CARET);
+	KillTimer(m_Window, TIMER_WRITE_OPTIONS);
+
+	m_WriteOptionsScheduled = false;
+	if (m_PendingWriteOptions != 0)
+	{
+		WriteOptions(m_PendingWriteOptions, true);
+		m_PendingWriteOptions = 0;
+	}
```
poiru: You could extract this block (including the `KillTimer`) into a `WriteDeferredOptions()` function. That would avoid duplicating this same pattern in `Skin::Deselect()`, `Skin::Refresh()`, and the `SetTimer` lambda.

poiru (changes requested):
Thanks, this is better! Left a few more comments.

Mitanshp5:
Thanks for the suggestions @poiru! I've updated the PR with the revised approach:
- Rebased onto latest `master`.
- Reverted all setter signatures and added `m_DeferOptionWrites` with a `Scoped` guard at the top of `Skin::DoBang()`.
- Moved debounce timer scheduling inside `Skin::WriteOptions()`.
- Extracted `WriteDeferredOptions()` to centralize timer cleanup and pending flushes across `Dispose()`, `Refresh()`, `Deselect()`, and the timer callback.

Ready for another look whenever you have time!

poiru:
Thank you! Manually merged as 053ddc6

